PR #64, opened 2015-03-23
Support for local, build-only extensions which are removed on deployment

vitorio:
Not expecting this to be merged as-is, but rather a starting point for a conversation.

Deploying to an object store like S3 means you don't have to stick with filesystem restrictions.  You can have an "about-us" URL without an extension and without a trailing slash, and also "about-us/image.gif" under it, because they're uniquely named objects, not files in a hierarchy.  See:

- http://vitorio-test-grow.s3-website-us-east-1.amazonaws.com/about-us
- http://vitorio-test-grow.s3-website-us-east-1.amazonaws.com/about-us/bench.gif

This works out of the box with the Grow development server, by changing the codelab's `content/pages/_blueprint.yaml`:

    path: /{slug}               # The URL path format.

And changing the codelab's main `podspec.yaml`:

    static_dirs:
      - static_dir: /static/
        serve_at: /static/
      - static_dir: /assets/about-us/
        serve_at: /about-us/

But, on build and deployment, it generates an "index.html" inside the "about-us" directory instead.  This is understandable, given the limitations of a filesystem, but what we really want is for "about-us" to live in the root as a file, not be a directory.

The code changes in this pull request add the concept of a build-only file extension that is removed upon deployment to a supported object store.  Adding this to `podspec.yaml` triggers the additional codepaths to generate "about-us.objstore.html" in the root, and then remove that `buildsuffix` when deploying:

    flags:
      buildsuffix: .objstore.html

I'm sure there's a better way to handle this, looking forward to your thoughts, feedback and critique.

### DIFF
--- a/grow/commands/build.py
+++ b/grow/commands/build.py
@@ -18,7 +18,7 @@ def build(pod_path, out_dir):
   pod.preprocess()
   try:
     if 'buildsuffix' in pod.flags:
-      paths_to_contents = pod.dump(suffix=pod.flags['buildsuffix'])
+      paths_to_contents = pod.dump(buildsuffix=pod.flags['buildsuffix'])
     else:
       paths_to_contents = pod.dump()
     repo = utils.get_git_repo(pod.root)
--- a/grow/commands/build.py
+++ b/grow/commands/build.py
@@ -17,8 +17,8 @@ def build(pod_path, out_dir):
   pod = pods.Pod(root, storage=storage.FileStorage)
   pod.preprocess()
   try:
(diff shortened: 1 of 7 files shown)


vitorio:
Hi, okay, after #64, it occurred to me that, if these temporary extensions are object store deployment-specific, and temporary, then they should probably have their scope limited to deployment, both in code and configuration, and the user shouldn't really have to deal with them otherwise.  Submitting a new pull request, #65, which I hope you'll consider.

PR #65, opened 2015-03-28
Object store-specific temporary deployment extensions

vitorio:
Hi, okay, after #64, it occurred to me that, if these temporary extensions are object store deployment-specific, and temporary, then they should probably have their scope limited to deployment, both in code and configuration, and the user shouldn't really have to deal with them otherwise.

As with #64, this pull request supports deploying resources to an object store (like S3) without extensions, by building on the filesystem with a temporary extension and then removing it on `write_file`, while also still supporting subresources, as seen here:

 - http://vitorio-test-grow.s3-website-us-east-1.amazonaws.com/about-us
 - http://vitorio-test-grow.s3-website-us-east-1.amazonaws.com/about-us/bench.gif

(That deployment is new, using this code, not using #64's code.)

This is already supported using the built in `grow run` development server with a `_blueprint.yaml` of:

    path: /{slug}               # The URL path format

And a `podspec.yaml` including:

    static_dirs:
      - static_dir: /static/
        serve_at: /static/
      - static_dir: /assets/about-us/
        serve_at: /about-us/

This new pull request enables a new configuration option only in the S3 deployment code:

    deployments:
      default:
        destination: local
        out_dir: my-codelab-launch/
      s3:
        destination: s3
        bucket: vitorio-test-grow
        extensionless_paths: yes

(By default, `extensionless_paths` is "no", and not needed to be specified at all.  There is also an option to specify the temporary extension used, but a reasonable default is also provided.)

No other deployment destinations are affected.  `extensionless_paths` is trapped appropriately in `deploy.py` for destinations which do not support it.  The `dump` function in `pods.py` is made more explicit to support both `index.html` and temporary extensions, and `suffix` is renamed to `main_page_suffix` to match `google_cloud_storage.py`.

I hope you'll consider accepting this pull request.  I'm also idling in IRC if you'd like to discuss it further there.

### DIFF
--- a/grow/commands/deploy.py
+++ b/grow/commands/deploy.py
@@ -35,7 +35,15 @@ def deploy(deployment_name, pod_path, build, confirm, test, test_only, auth):
     if test_only:
       deployment.test()
       return
-    paths_to_contents = pod.dump()
+    build_extension = ''
+    try:
+      extensionless_paths = deployment.config.field_by_name('extensionless_paths')
+    except KeyError:
+      pass
+    else:
+      if deployment.config.extensionless_paths:
+        build_extension = deployment.config.build_extension
+    paths_to_contents = pod.dump(build_extension=build_extension)
     repo = utils.get_git_repo(pod.root)
     stats_obj = stats.Stats(pod, paths_to_contents=paths_to_contents)
     deployment.deploy(paths_to_contents, stats=stats_obj, repo=repo,
--- a/grow/deployments/destinations/amazon_s3.py
+++ b/grow/deployments/destinations/amazon_s3.py
@@ -17,6 +17,8 @@ class Config(messages.Message):
   access_secret = messages.StringField(3)
   env = messages.MessageField(env.EnvConfig, 4)
   keep_control_dir = messages.BooleanField(5, default=False)
+  extensionless_paths = messages.BooleanField(6, default=False)
+  build_extension = messages.StringField(7, default='.objstore.html')
 
 
 class AmazonS3Destination(base.BaseDestination):
@@ -46,6 +48,9 @@ class AmazonS3Destination(base.BaseDestination):
     return self.write_file(path, content, policy='private')
 
   def read_file(self, path):
+    if self.config.extensionless_paths:
+      if path.endswith(self.config.build_extension):
+        path = path[:-len(self.config.build_extension)]
     file_key = key.Key(self.bucket)
     file_key.key = path
     try:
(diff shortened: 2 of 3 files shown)


jeremydw:
I was actually going to discuss the very same thing with you – that there may not be a need to actually write files with a temporary suffix in the build directory, if indeed they should be temporary. Glad to see we're on the same page here!

I had a busy week, but will review this weekend. Thank you!

jeremydw:
Looks good overall! I'm still testing out the APIs a bit (specifically, the Google Cloud Storage API) to see if there's a way we can achieve the same behavior through an S3/GCS API configuration, or if there's existing terminology we could use for the configuration format so we can be internally consistent with the object store.

As I'm experimenting, it'd also help to understand the use case a little more. When a bucket is configured with "website configuration" enabled, the behavior is similar to your staging site but slightly different. I'm guessing you would like this feature so that objects are *never* served at "/index.html" – is that right? Or is there something else I'm missing?

As an aside, I noticed that requests to http://vitorio-test-grow.s3-website-us-east-1.amazonaws.com/about-us/ (note the trailing slash) 404, whereas if website configuration were enabled, they would redirect to the resource.

Sorry for the delay and thanks!

vitorio:
You always need to specify an "index document," because you need one for the site root `/`, and having that be named `index.html` is fine with me, it's as standard or expected as anything.  The current patch continues to use `index.html` if you have a trailing slash on your slug or path.  I just don't want to have to have a trailing slash, when trailing slashes and index documents are really a kludge for filesystem-based stores.  If this were a REST API, for example, you'd expect `http://example.com/api/endpoint` to be a single resource, and `http://example.com/api/endpoint/` to be a collection of other things.

Here's a proper example where "website configuration" is enabled:

- http://vitor.io/genuine-outbreaks-of-the-future is an extensionless resource.
- http://vitor.io/genuine-outbreaks-of-the-future/ is a "folder" which 403s, because there's no index.html.  I think this is appropriate: if you think of this as a collection of sub-resources, I don't want you poking around.
- http://vitor.io/genuine-outbreaks-of-the-future/27-7.jpg is a sub-resource under that first extensionless resource.
- http://vitor.io.s3-website-us-east-1.amazonaws.com/genuine-outbreaks-of-the-future is the equivalent S3-named endpoint.

If I put an `index.html` file in the folder, then it serves that up when the trailing slash is present, and serves up the original extensionless resource when it's not.

Does that help?

This temporary-extension-and-then-rename is how I manage the process manually now.  I haven't looked into how Google Cloud Storage serves up files, but I can, if it would help.

vitorio:
Oh, apparently the folder 403s because I don't have an error document configured.  If I configure an error document (like `404.html` in codelab), then it 404s.

It doesn't seem like I can create a redirection rule to make URLs with trailing slashes redirect to URLs without trailing slashes.

jeremydw:
Hi again! I've been doing a lot of thinking about this, and I came up with an alternative that hides the "build extension" from the user – but does allow them to specify "trailing slash redirect" behavior for leaf resources (which I think is the core of this PR).

Please take a look here and let me know what you think –
https://github.com/grow/pygrow/pull/66

vitorio:
Closing in anticipation of #66 being merged

PR #74, opened 2015-05-13
Ignore hidden docs in search_docs().

stevenle:


stevenle:
Hmm, actually hang on, this still isn't working.

stevenle:
Okay, actually, this does work. Had a local build issue.

PR #103, opened 2015-08-13
Update README pip install gotchas for Linux.

fotinakis:
@jeremydw for review

Installing grow from pip on a blank Ubuntu install fails on two dependencies:
screen shot 2015-08-13 at 12 31 06 pm
screen shot 2015-08-13 at 12 39 49 pm

Unfortunately you have to upgrade the "six" package independent of the other steps.

Also, pip apparently only has 0.0.52, not the latest 0.0.53 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,13 +63,6 @@ Once installed, you can...
     ./scripts/test                                 # Runs tests.
     ./scripts/grow                                 # Runs `grow` command line program.
 
-#### Gotchas
-
-From a fresh system, you may need a few things to build a Grow release from scratch:
-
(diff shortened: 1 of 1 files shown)


jeremydw:
That looks good, but those instructions were supposed to only be for building the packed releases (with the `./scripts/build_release` script). Is plain old `pip install grow` failing for you too? If so we should fix the issue. Actually it looks like it's indeed failing on Travis: https://travis-ci.org/grow/pygrow/builds/75525424

fotinakis:
Ah interesting — yes, `pip install grow` fails on a vanilla Ubuntu install (in Docker) with the above errors. I believe these updates fix it, and we could apply them to the travis config as well.

PR #161, opened 2015-12-09
Add semantic version enforcement

chargrizzle:
This PR enforces the podspec version requirement if it is specified


jeremydw:
Looks great. Thank you!

Issue #55, opened 2014-11-15
Custom preprocessors / plugins


scisci:
Would be great to be able to write custom preprocessor in python, or jinja extensions, without having to fork the repo. If there was some way to tell grow to add plugins dynamically, would make it much more customizable.

stucox:
The Grow preprocessor class design already suits defining custom processors with whatever parameters the author might want to expose. And of course Jinja extensions are well-defined. Just need a way of telling Grow to import & use them.

Maybe this could start as simple as a `extensions` definition in podspec.yaml, e.g.:

```yaml
extensions:
- extensions.jinja2.ForceEscape
- extensions.preprocessors.ClosureBuilder
```

Grow would just need to:
* Find these classes relative to the pod root (i.e. in `/extensions/jinja2.py` and `/extensions/preprocessors.py` respectively)
* Determine whether each is a Jinja2 extension or Grow preprocessor based on whether they inherit from `jinja2.ext.Extension` or `grow.pods.preprocessors.base.BasePreprocessor` respectively
* Add Jinja extensions to the `extensions` kwarg when creating the Jinja environment
* Call `register_preprocessor()` for each Grow preprocessor

`extensions/` would not need to be a builtin directory name — it’s just probably how I’d structure things in my own projects.

stucox:
Worth noting authors would need to bundle custom Jinja filters into an extension, or Grow would need to provide special syntax for loading them.

jeremydw:
As of January, Grow has supported an extension system, allowing authors to write custom Jinja extensions and custom preprocessors. This, with the ability to pair Grow with Gulp, should hopefully be sufficient here!

http://grow.io/docs/extensions/
http://grow.io/docs/preprocessors/#gulp

WillsB3:
Sorry to comment on an old issue, but not sure where better to ask this question...

I'm experimenting with writing a custom Jinja tag for use in a Grow project. This tag will take an image path and transform it from a regular local image URL to a [proxied URL](https://github.com/imgix/imgix-python#basic-usage) for the image on the image CDN service [imgix](https://www.imgix.com/). However I haven't been able to figure out how to register the tag (e.g. do the equivalent of `environment.filters['triplicate'] = do_triplicate` but for a tag instead of a filter) - can anyone provide any pointers on how to achieve this?

[...]

PR #169, opened 2016-01-18
Add support for Jinja2 extensions and custom preprocessors

stucox:
Simple solution for #55. A slight variation on what I proposed there: decided it was easier & more explicit to separate jinja2 / preprocessors in podspec.yaml:

```yaml
extensions:
  jinja2:
  - path.to.CustomJinja2Extension
  preprocessors:
  - path.to.CustomPreprocessor
```

[See the example in grow/testing/testdata/pod/podspec.yaml](https://github.com/stucox/pygrow/blob/feature/extensions/grow/testing/testdata/pod/podspec.yaml#L54).

Custom Jinja2 filters would need to be wrapped in an extension class (i.e. a subclass of) — the tests I’ve written for this include an example of this. Could add syntax to allow filters to be declared directly in podspec.yaml (then added to the env via the `filters` arg) — maybe at a later date?

Happy to write some docs for this once merged — should that be a PR against https://github.com/grow/growsdk.org?


jeremydw:
Great work on this and thank you for the contribution. :) The solution looks great from a quick glance.

Before merging, I'm doing a tiny bit of research on naming and the syntax in podspec.yaml. Can you adjust the PR to be based off of `develop` instead of `master`? `master` is what's currently released.

Also, before releasing this I'd like to perhaps iterate on the "public API" for writing extensions -- for example, it might be cleaner to allow `import grow` and then for the base class to be `grow.Preprocessor` (via an alias in the right place in `__init__.py`). This would basically be the first time the internal paths of the Python program have any implication on how folks construct sites.

Additionally (just as a reminder to myself) – need to test the behavior of the imports in the frozen/binary environment.

stucox:
Closing this — will re-raise against `develop`.

PR #171, opened 2016-02-02
Changed messages associated with 'grow run' command to logging messages

codedcolors:
I've updated a few things that were previously being printed to use logging instead, so that these messages are passed to stdout. 

In my specific use case, I'm using a gulp task to run the 'grow run' command via subprocess, and the `print` messages aren't passed into the stdout, so the task can't display the `print_server_ready_message` or 'check_for_sdk_updates` message. This change enables me to pass these notifications to the subprocess and display them along with the gulp output.


jeremydw:
Lovely. Thanks @codedcolors !

PR #217, opened 2016-05-17
Remove side effects on cloud storage deployment

jeremydw:
@vitorio PTAL


vitorio:
This matches what I have on the S3 side, and probably a good call to not change any versioning settings.  I don't have a GCS account to test that side, though.

PR #218, opened 2016-05-18
small fix for getting subprocess resolving executables on windows

jamesvecore:
This fixes an issue with grow trying run the gulp preprocessor on windows machines.

### DIFF
--- a/grow/common/sdk_utils.py
+++ b/grow/common/sdk_utils.py
@@ -92,12 +92,15 @@ def check_for_sdk_updates(auto_update_prompt=False):
 def get_popen_args(pod):
     node_modules_path = os.path.join(pod.root, 'node_modules', '.bin')
     path = os.environ['PATH'] + ':{}'.format(node_modules_path)
-    return {
+    args = {
         'cwd': pod.root,
         'env': {
             'PATH': path,
         },
     }
+    if os.name == 'nt':
+        args["shell"] = True
+    return args
 
 
 def install(pod, gerrit=None):
--- a/grow/common/sdk_utils.py
+++ b/grow/common/sdk_utils.py
@@ -91,12 +91,11 @@ def check_for_sdk_updates(auto_update_prompt=False):
 
 def get_popen_args(pod):
     node_modules_path = os.path.join(pod.root, 'node_modules', '.bin')
-    path = os.environ['PATH'] + ':{}'.format(node_modules_path)
+    env = os.environ.copy()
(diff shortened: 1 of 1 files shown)


jamesvecore:
This only fixes the preprocessor subprocess execution on windows. There are a number of path related issues that come up when trying to build an actual site on windows that are not addressed by this pull request.

jeremydw:
Hi James -- Thanks for this. We've been trying to get past this without `shell=True` so far. Can you elaborate why `shell=True` is needed for Windows? Maybe there's just an env issue that can be amended?

If we do indeed need `shell=True`, I'm inclined to just enabling it for all platforms.

jamesvecore:
For the gulp preprocessor specifically, gulp.js is executed by node on windows through a file called gulp.cmd which is a batch file. When shell=True is not passed, subprocess.Popen() will not execute .cmd batch files on windows. Windows doesn't support shell script sh-bangs so often unix-y tools are launched with .cmd wrapper batch files with the same root name. This true for git, most node commands (including npm), and a lot of python module support scripts. In fact grow itself will not run from the cmdline on windows after module installation because it does not provide a wrapper grow.cmd that would launch the python interpreter targeting the 'grow' script in the python scripts folder. 

On windows, it is also necessary to pass a full set of environment variables in order for the shell execution to work. The current version of grow passes an env object with only 'PATH' set and this causes windows to fail to execute .cmd batch files with a random negative exit code.

One final issue is that when setting the environment for subprocess on windows, all the values must be str, as unicode is not supported (hence the str() conversion call).

That said, I'm finding a number of other issues preventing me from running a real grow project on windows. So this patch as it stands won't help much for windows people. As time permits, I will try to track these down and add them to this pull request (include a fix for bootstrapping the grow command itself on windows).

jeremydw:
Gotcha -- thanks for these details. Would you be able to rebase this PR against `develop` -- then we can merge it on in? (Currently PRs should go against `develop`, though we're going to change it to be against `master` like most projects soon, see discussion from: https://github.com/grow/grow/issues/105#issuecomment-217340062)

I agree with setting `shell=True` for Windows based on your explanation. I'd also like to fully support Windows, it just hasn't been a requirement thusfar. We'd welcome any other changes needed to get this going on Windows.

jeremydw:
Merged this in here to `master`. Thanks!

https://github.com/grow/grow/commit/78908471b79760cb0a7b8643c4b2229f2f33e291

Issue #157, opened 2015-11-23
Grow extract extracts strings for non-specified locales


matthewhall:
Looks like running `grow extract --localized` will extract all strings from a YAML for all available locales, regardless of whether they're only overridden for a particular locale or not.

For example:

```
some_key@: "Some text"

---
$locales:
  - en_au
  - en_nz

some_key@: "Some different text"
```

Running `grow extract --localized` will extract "Some different text" for all locales not just `en_au` and `en_nz`.

Thanks.

jeremydw:
This should be fixed in `master`, now and we've got a larger tracking issue that should also be all fixed up in #204. Thanks a bunch.

Issue #206, opened 2016-05-03
Disallow document parts without locales declared (maybe all overrides?)


stucox:
Currently, if you have the following HTML / Markdown document:

```yaml
---
# front matter 1
---
Main body
---
# front matter 2
---
New body
---
```

“New body” will silently override “Main body” for all locales. I think this should throw an error — as there’s no obvious reason why someone would intentionally do this.

Similarly:

```yaml
---
$localization:
  locales:
    - en
    - fr
---
$locale: en
foo: 1
---
foo: 2
---
```

`foo` will equal 2 for both `en` and `fr` locales here. The chances are this would be a mistake: if the developer meant for `foo` to equal 2 for just `fr`, he/she should have put that in the main document front matter.

---

## Slightly more contentious cases:

```yaml
---
$localization:
  locales:
    - en
    - fr
---
$locale: en
foo: 1
---
$locale: en
foo: 2
---
```

I’d like to see an error here too. There’s clearly redundant code.

Although I know some Grow users would like more complex multiple-overrides to be possible, e.g.:

```yaml
---
$localization:
  locales:
    - en
    - fr
    - de
foo: 1
bar: 1
---
$locales:
  - en
  - fr
foo: 2
bar: 2
---
$locale: en
foo: 3
---
```

I believe @willsb3 and co would like this to result in `foo = 3` and `bar = 2` for `en`, `foo = 2` and `bar = 2` for `fr`, etc: i.e. values would be merged, so this can be used as a method of managing complex variation between locales without repetition…… but I’d also like to see an error in this case, because:

- It would be consistent with throwing an error for the previous (clearly redundant code) case
- I favour composition over inheritance, so would rather see common values defined in the top front matter & included into locale-specific sections without overriding

---

Hidden agenda: working on #204, if cases like this were disallowed, it also gives fewer cases which need to be defined/covered/tested… 😄

jeremydw:
@meizon can you comment on the complex multi-override that @stucox laid out here? Prior to a commit I made based on some work I did with you, Grow would override the base document with any localized documents of the same locale. I think both ways make sense but I also think the behavior we had prior to that change was simpler and less hairy.

I'm in favor of documents throwing an error when multiple sets of front matter are defined without locales -- thanks for raising that. Even though I personally prefer the behavior that does not permit overriding the base locale as it's more straightforward I also understand the use case @willsb3 etc. follows as it avoids repetition so it's probably worth a little more discussion with them.

[...]

PR #231, opened 2016-06-16
Refactor `Catalogs.extract()` to implement more consistent behaviour

### LINKED ISSUE #204: Clarify `grow extract --localized` behaviour
cc @jeremydw, @willsb3 & @matthewhall 

In attempting to fix #157, I found various other cases of `grow extract --localized` which didn’t act as I’d expect.

These all relate to strings being extracted for locales for which they aren’t needed, which:

(a) Makes `grow filter` report translations are missing when they aren’t, hence making it harder to identify translations which _are genuinely_ missing
(b) Creates unnecessary work for translators

The behaviour of `grow extract --localized` isn’t documented on grow.io and doesn’t have very extensive test coverage. It’d be good to clarify exactly which tagged strings in which context should be expected to be extracted into which locales’ message catalogs, get tests in place, and refactor (hopefully simplify) `Catalogs.extract()` to match.

My effort so far:
* In https://github.com/stucox/pygrow/commit/16137fdaae394e47e5af1a330822b3abf0ce248c I increased the test coverage of `Catalogs.extract()`, describing the **current behaviour** (all tests pass on this commit)
* Those cases from [L91](https://github.com/stucox/pygrow/commit/16137fdaae394e47e5af1a330822b3abf0ce248c#diff-d1b59363637d70582c52f59100e5e256R91) onwards represent behaviour **I think is undesirable**
* In https://github.com/stucox/pygrow/commit/9cd889381a9d7e05fa8e5fb85731aa1c067343c5 I changed these to represent the behaviour I think **should be considered correct** — but I think it needs some discussion
* I also attempted to _just_ fix #157 without changing other behaviour, but my “clean” solution would have required many special cases to retain the behaviour I believe is undesirable

These show that in many contexts locale configuration is ignored by `grow extract --localized` and strings are extracted for all locales mentioned anywhere in the pod (not even just those in podspec.yaml) and it’s very hard to ensure strings only appear in the relevant message catalogs.

---

I propose we standardize this as follows:

1. The `/content/` folder is considered a tree of scopes (_pod_ > _collection_ > _document_ > _document part_), such that:
   - Tagged strings are only extracted for the locales declared for the current scope
   - If no locales are declared for the current scope, it assumes the locales of its parent scope
   - YAML, Markdown, HTML and CSV files are all considered “documents” for this purpose and follow the same rule for determining relevant locales
   - Ergo, files with no front matter (including CSVs, which never have front matter) will obviously always assume the collection’s locales (or the pod’s, if no locales are defined in the collection’s blueprint)
2. Tagged strings in `/views/` are simply extracted for all locales named in podspec.yaml

Rule 2 could alternatively state _“all locales named anywhere in the pod”_… there are arguments for/against each, but keen to get general thoughts on the above before launching into that debate.

I’ve omitted `/data/` as I’m not convinced treating it as a special folder is necessary: that isn’t documented anywhere, and it could be argued a `/content/data/` collection (which according to the rules above can also contain translatable CSV files) is more powerful, or alternatively `!g.csv <>` syntax can already be used to bring CSV data into a document (a translatable context).

stucox:
Fixes #204, #157 & #206.

Pretty much by definition this is a breaking change — see #204 and the aim & the included tests for details of cases I’ve covered. Not sure the best way to communicate this to users.

Note that in this version `/data/` **isn’t** treated as a special directory anymore.

I’ve tested this on a very large project I work on and it does what I’d expect:
- Removes unnecessary translations (because of #157)
- No longer extracts tagged terms in ‘global’ locations (i.e. where no document part, document or collection locale can be determined) for all locales defined in the pod: just for those locales in podspec.yaml

This last point caused some work for us: we had a “globals.yaml” file which assumed everything was relevant to _every_ locale (which was wrong: not every locale used those) and hence didn’t declare any locales… which worked until now. Solved by either listing all relevant locales in that file, or listing them all in podspec.yaml (the list there only contained _primary_ locales, which might be better defined at collection level).

**Note:** Would appreciate if people could test on some other projects.

I’ll try a few others of ours while @jeremydw is checking out the code.

### DIFF
--- a/grow/pods/catalog_holder.py
+++ b/grow/pods/catalog_holder.py
@@ -235,7 +235,7 @@ class Catalogs(object):
 
         # Extract from root of /content/:
         for path in self.pod.list_dir('/content/', recursive=False):
-            if path.endswith('.yaml') or path.endswith('.yml'):
+            if path.endswith(('.yaml', '.yml')):
                 pod_path = os.path.join('/content/', path)
                 self.pod.logger.info('Extracting: {}'.format(pod_path))
                 utils.walk(
--- a/grow/pods/catalog_holder.py
+++ b/grow/pods/catalog_holder.py
@@ -189,9 +189,18 @@ class Catalogs(object):
         # Extract from collections in /content/:
         # Strings only extracted for relevant locales, determined by locale
         # scope (pod > collection > document > document part)
+        last_pod_path = None
         for collection in self.pod.list_collections():
             for doc in collection.list_docs(include_hidden=True):
-                self.pod.logger.info('Extracting: {}'.format(doc.pod_path))
+                if doc.pod_path != last_pod_path:
+                    self.pod.logger.info(
+                        'Extracting: {} ({} locale{})'.format(
+                            doc.pod_path,
+                            len(doc.locales),
+                            's' if len(doc.locales) > 1 else '',
+                        )
+                    )
+                    last_pod_path = doc.pod_path
 
                 # If doc.locale is set, this is a doc part: only extract for
                 # its own locales (not those of base doc).
--- a/grow/pods/catalog_holder.py
+++ b/grow/pods/catalog_holder.py
@@ -231,7 +231,7 @@ class Catalogs(object):
                     rows = self.pod.read_csv(pod_path)
                     for i, row in enumerate(rows):
                         for key, msgid in row.iteritems():
-                            _handle_field(pod_path, collection.list_locales(), msgid, key, row)
+                            _handle_field(pod_path, collection.locales, msgid, key, row)
 
         # Extract from root of /content/:
         for path in self.pod.list_dir('/content/', recursive=False):
--- a/grow/pods/catalog_holder.py
+++ b/grow/pods/catalog_holder.py
@@ -10,6 +10,7 @@ from grow.pods import messages
 import click
 import collections
 import os
+import StringIO
 import tokenize
 
 
@@ -111,42 +112,18 @@ class Catalogs(object):
         if content:
             importer.import_content(content=content, locale=locale)
 
-    def _get_or_create_catalog(self, template_path):
-        exists = True
-        if not self.pod.file_exists(template_path):
-            self.pod.create_file(template_path, None)
-            exists = False
-        catalog = pofile.read_po(self.pod.open_file(template_path))
-        return catalog, exists
-
-    def _add_message(self, catalog, message):
-        lineno, string, comments, context = message
-        flags = set()
-        if string in catalog:
-            existing_message = catalog.get(string)
-            flags = existing_message.flags
-        return catalog.add(string, None, auto_comments=comments, context=context,
-                           flags=flags)
-
-    def _should_extract_as_csv(self, given_paths, path):
-        ext = os.path.splitext(path)[-1]
-        if ext != '.csv':
-            return False
-        return given_paths is None or path in given_paths
-
-    def _should_extract_as_babel(self, given_paths, path):
-        if os.path.splitext(path)[-1] not in _TRANSLATABLE_EXTENSIONS:
-            return False
-        return not given_paths or path in given_paths
-
     def extract(self, include_obsolete=False, localized=False, paths=None,
                 include_header=False, locales=None, use_fuzzy_matching=False):
         env = self.pod.create_template_env()
 
-        all_locales = set(list(self.pod.list_locales()))
-        message_ids_to_messages = {}
-        paths_to_messages = collections.defaultdict(set)
-        paths_to_locales = collections.defaultdict(set)
+        # {
+        #    locale1: locale1_catalog,
+        #    locale2: locale2_catalog,
+        #    ...
+        # }
+        # This is built up as we extract
+        localized_catalogs = {}
+        unlocalized_catalog = catalogs.Catalog()  # for localized=False case
 
         comment_tags = [
             ':',
@@ -156,37 +133,24 @@ class Catalogs(object):
             'silent': 'false',
         }
 
-        # Extract from content files.
-        def callback(doc, item, key, unused_node):
-            # Verify that the fields we're extracting are fields for a document
-            # that's in the default locale. If not, skip the document.
-            _handle_field(doc.pod_path, item, key, unused_node)
+        def _add_to_catalog(message, locales):
+            # Add to all relevant catalogs
+            for locale in locales:
+                if locale not in localized_catalogs:
+                    # Start with a new catalog so we can track what's obsolete:
+                    # we'll merge it with existing translations later.
+                    # *NOT* setting `locale` kwarg here b/c that will load existing
+                    # translations.
+                    localized_catalogs[locale] = catalogs.Catalog(pod=self.pod)
+                localized_catalogs[locale][message.id] = message
+            unlocalized_catalog[message.id] = message
 
-        def _add_existing_message(msgid, locations, auto_comments=None,
-                                  context=None, path=None):
-            existing_message = message_ids_to_messages.get(msgid)
-            auto_comments = [] if auto_comments is None else auto_comments
-            if existing_message:
-                message_ids_to_messages[msgid].locations.extend(locations)
-                paths_to_messages[path].add(existing_message)
-            else:
-                message = catalog.Message(
-                    msgid,
-                    None,
-                    auto_comments=auto_comments,
-                    context=context,
-                    locations=locations)
-                paths_to_messages[path].add(message)
-                message_ids_to_messages[message.id] = message
-
-        def _handle_field(path, item, key, node):
+        def _handle_field(path, locales, msgid, key, node):
             if (not key
                     or not key.endswith('@')
-                    or not isinstance(item, basestring)):
+                    or not isinstance(msgid, basestring)):
                 return
-            # Support gettext "extracted comments" on tagged fields. This is
-            # consistent with extracted comments in templates, which follow
-            # the format "{#: Extracted comment. #}". An example:
+            # Support gettext "extracted comments" on tagged fields:
             #   field@: Message.
             #   field@#: Extracted comment for field@.
             auto_comments = []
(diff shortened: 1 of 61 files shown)


jeremydw:
Awesome, thank you for putting all the time into the change and also for the excellent test coverage. I'll start the review with some style-related stuff just to keep consistency with the rest of the project and then I'll dig in after to see if I have any comments on the overall design. I gave a cursory look at it yesterday and it looks pretty great.

grow/pods/catalog_holder.py
```diff
@@ -156,177 +133,165 @@ def extract(self, include_obsolete=False, localized=False, paths=None,
...
-
-        # Localized message catalogs.
+                    rows = self.pod.read_csv(pod_path)
+                    for i, row in enumerate(rows):
+                        for key, msgid in row.iteritems():
+                            _handle_field(pod_path, collection.locales, msgid, key, row)
+
+        # Extract from root of /content/:
+        for path in self.pod.list_dir('/content/', recursive=False):
+            if path.endswith('.yaml') or path.endswith('.yml'):
```
jeremydw: `endswith` accepts tuples, so you can do `path.endswith(('.yaml', '.yml'))`

grow/pods/storage/file_storage.py
```diff
@@ -33,12 +33,15 @@ def stat(filename):
         return os.stat(filename)
 
     @staticmethod
-    def listdir(dirpath):
+    def listdir(dirpath, recursive=True):
```
jeremydw: Thanks for adding this. :)

grow/testing/testing.py
```diff
@@ -7,16 +7,17 @@
 TESTDATA_DIR = os.path.abspath(os.path.join(_here, 'testdata'))
 
 
-def create_test_pod_dir():
+# pod_dir is relative to testdata/ dir
+def create_test_pod_dir(pod_dir='pod'):
```
jeremydw: The term we use for what you're calling `pod_dir` is `root`. (See an example here: https://github.com/grow/grow/blob/master/grow/commands/run.py#L33). Just for consistency can you switch `pod_dir` to `root` here?

jeremydw:
In response to this comment, "No longer extracts tagged terms in ‘global’ locations", does that behavior only kick in when `--localized` is used with `grow extract`? In other words, are there no changes if `--localized` is not specified?

grow/pods/catalog_holder.py
stucox: This I didn’t know! Will update.

grow/testing/testing.py
stucox: Yep

stucox:
> In response to this comment, "No longer extracts tagged terms in ‘global’ locations", does that behavior only kick in when --localized is used with grow extract? In other words, are there no changes if --localized is not specified?

Well, sorta N/A: `grow extract` (without `--localized`) always ignored ‘locale contexts’* of tagged strings (i.e. the `locales` defn of the doc, collection or pod the string is in) and just stuck anything marked for translation in `/content/`, `/views/` and `/data/` in `messages.pot` — so the only change there is `/data/` is now ignored. Maybe this could do with some more test coverage.

_* I like this term. I might use it more._

jeremydw:
Makes sense, thanks for explaining. The term "locale context" sounds good. ;)

I'm OK with the breaking change of ignoring `/data/` as we're aligned in moving forward with simply putting all content (partials, data files, etc.) in the `/content/` directory anyway. `/data/` was special and undocumented so I'm OK breaking any behavior there.

ALSO – This should definitely be a separate feature/PR but I'd also like to add a new configuration option to `localization` in `podspec.yaml` where the user can specify the defaults for the `extract` command.

```
localization:
  locales: [...]
  default_locale: [...]
  extract:
    localized: true  # Defaults `--localized` when `grow extract` is run.
```

That would simplify translation management across multiple developers and help with any continuous translation initiative we work on.

stucox:
You can still put stuff in `/data/` — but if you want it extracted for localization, you need to `!g.yaml()` / `!g.csv()` it into a document in `/content/`. That could probably do with test coverage too.

stucox:
Agree that config option would be nice — but low priority IMO. Kinda like teams to know what the commands they’re running are doing.

jeremydw:
Ha, well, if we provide some built-in optional commit hooks (such as a pre-commit `grow extract` followed by a diff/report and a post-commit `grow stage`) it'll be important for the commands to be self-configured.

stucox:
👍  We’ve played with an “feature finished” script too, which runs linting, `grow extract --localized` (and complains if your translations files are out of date), `grow stage`, pushes to our code review tool & sticks changes a ticket status in our bug tracker. So would be helpful for that too.

(we decided pre/post-commit was a bit much)

jeremydw:
Just a note: I ran this against against a project I work on and everything works as intended! I noticed this exposes the reason behind that `use_podspec_locales` identifier we chatted about on email.

As now intended, in absence of that identifier, and in absence of any `$localization` config,  `grow extract --localized` no longer extracts these partials. Here's an example file that I'm working with. Note that it's a partial/includable content file and therefore doesn't have a URL mapped to it.

```
---
foo: bar
---
$locale: de
foo: baz
```

Also, there is no `$localization` in the document or in its blueprint.

I think the change in behavior introduced by this PR is correct -- to not extract this content at all. To get the previous behavior, you'd want the equivalent of `use_podspec_locales` but that's wordy and just weird, so...

After merging this PR in, I'm thinking about making a change to the default behavior of these partial files that do not have a `$localization` key and want your thoughts, here are my proposals:

1. Accept `$localization: ~` (null character) as an indicator that the file should be localized. This would be the equivalent of `use_podspec_locales`, it just leverages consistent syntax. Since this file is a partial and does not have a path mapped to it, it wouldn't need `$localization: path: ...` and therefore can just accept `$localization: ~` as the "I am localized too" flag.
1. In lieu of a `$localization` key, if the user has specified a sub-document in the file with `$locale: foo`, then that should automatically imply `$localization: ~`. Note that this would not change the current behavior of inheriting the podspec locales, it would just automatically indicate that the doc is localized.

Therefore, in order to get the previous behavior back and have this file extracted too, I would have to do nothing (since `$locale: de` would imply `$localization`) -- or, if the partial didn't have `$locale: de`, I would simply have to add `$localization: ~` to indicate that I want the file to be localized and extracted to all locales.

Thoughts on that?

stucox:
I'm pretty sure with the new implementation strings in that file will still be extracted — but it looks like they're not tagged for translation?

If it were this:

```yaml
---
foo@: bar
---
$locale: de
foo@: baz
```

I'd expect `bar` to be extracted for podspec locales and `baz` to be extracted for `de` only.

grow/pods/catalogs.py
```diff
@@ -118,13 +122,18 @@ def init(self, template_path, include_header=False):
         self.fuzzy = False
         self.save(include_header=include_header)
 
-    def update_using_catalog(self, catalog_to_merge, use_fuzzy_matching=False):
+    def update_using_catalog(self, catalog_to_merge, use_fuzzy_matching=False,
+                             include_obsolete=False):
         super(Catalog, self).update(
             catalog_to_merge, no_fuzzy_matching=(not use_fuzzy_matching))
+        # Don't use gettext's obsolete functionality as it polutes files: merge
```
stucox: @jeremydw Meant to ask — do you agree with this? Looks like the current implementation doesn't use gettext's functionality for obsolete strings, so I avoided it too: if you `include_obsolete` it'll just include obsolete terms as if they weren't obsolete. Seemed safest, in case obsolete syntax isn't widely supported by translation tools (dunno if this is the case, just a guess it might not be!)

jeremydw:
Bad explanation on my part, the strings are indeed tagged, e.g. `foo@: bar`.

You can take a peek but the reason they're not extracted is because without the `$localization` key Grow thinks the document isn't localized (believing it to just be of the default locale) and thus `doc_locales` (L208 on `catalog_holder.py`) contains just `en_US`.

This logging statement shows up when running `grow extract --localized`: `Extracting: /content/includes/footer.yaml (0 locale)`

As I mentioned I think this is correct behavior as-is (aside from the logging statement). The presence of `$localization: ~` or just a change to make `$localization` inferred if a document has `$locale: [...]` specified inside a part would fix the behavior.

grow/pods/catalogs.py
jeremydw: Oh, maybe. I see what you've done, but I don't know how the default functionality pollutes files other than leaving the terms in. What does it do exactly that's bad? I think obsolete terms in PO files are just comments preceded with `#~`. IMO if we could support that default obsolete behavior/syntax (when `--include-obsolete` is used) that'd be good? Feel free to make that outside the scope of this PR though.

grow/pods/catalogs.py
```diff
@@ -135,9 +144,17 @@ def update(self, template_path=None, use_fuzzy_matching=False,
...
-                  include_header=include_header)
+        # Don't use gettext's obsolete functionality as it polutes files: merge
+        # into main translations if anything
+        if include_obsolete:
+            self.merge_obsolete()
+        self.obsolete = odict()
+        self.save(include_header=include_header)
+
+    def merge_obsolete(self):
+        """ Copy obsolete terms into the main catalog """
```
jeremydw: Just a tiny style nit, but docstrings should end in punctuation and not be padded with space characters. ;) This isn't a blocker but I just wanted to raise this for consistency for future code you write!

stucox:
Yep that's a bug then — my intention was for `$localization` to be inferred if omitted. Will add a test for case and fix it.

grow/pods/catalogs.py
stucox: Raised #232 to discuss. Will leave as-is on this PR.

grow/pods/catalogs.py
stucox: ACK

jeremydw:
Thanks for adding those extra tests, now it looks like they're (correctly) failing and that matches up with the behavior that I reported when using `grow extract --localized` against my project.

Other than that I think we're good with all these changes. If you're happy with this I think we can merge it in as soon as the fix to infer `$localization` is added.

I *think* the change for that should be made here (https://github.com/stucox/pygrow/blob/c8bb8ab51de401459816db141bbb40ae5da17a8f/grow/pods/documents.py#L260).

With some pseudo code like:
- If `$localization` is not in fields, check whether document is split.
- To check whether a document is split, we can probably update `_handle_pairs_of_parts_and_bodies` in https://github.com/stucox/pygrow/blob/c8bb8ab51de401459816db141bbb40ae5da17a8f/grow/pods/formats.py#L124 to set some property on the `Format` instance (such as `has_localized_parts`).
- I *think* that if `$localization: locales: [...]` is *not* set on the base part, the list of locales should still be inherited from the collection's locales.

That'd just be one way to do it. Am open to another way or a simpler way of inferring whether a document is localized. LMK if you have any questions or want to pass this over to me for merging.

stucox:
Yep I'll fix those cases tomorrow. Thanks for the pseudocode.

jeremydw:
Merged by https://github.com/grow/grow/pull/239

PR #244, opened 2016-06-29
Revert gcs-oauth2-boto-plugin back to 1.14

stevenle:
gcs-oauth2-boto-plugin v1.8 is incompatible with oauth2client 2.0 or greater.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ click==4.0
 coverage==3.7.1
 dulwich==0.13.0
 ez-setup==0.9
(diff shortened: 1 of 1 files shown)


jeremydw:
All right -- let's revert this for now. I filed https://github.com/GoogleCloudPlatform/gcs-oauth2-boto-plugin/issues/25 against gcs-oauth2-boto-plugin earlier today to get them to update the release on PyPi which would be compatible with oauth2client >2.0.

PR #259, opened 2016-07-16
Allow document-level sitemap controls

jeremydw:
Thanks @stucox -- here's a feature that lets you use `$sitemap: enabled: false` on documents to disable inclusion in the sitemap.

As an FYI, it's not `$sitemap: false` since you can specify other things like `$sitemap.changefreq` and `$sitemap.priority`, plus `$sitemap: enabled: true` is consistent with `sitemap: enabled: true` in the podspec. Let me know if you think we should overload `$sitemap` though.

Fixes https://github.com/grow/grow/issues/258


stucox:
LGTM

Issue #265, opened 2016-08-02
Empty blueprint doesn’t inherit locales from podspec


stucox:
Since #239, I’d expect an empty (or even missing?) `_blueprint.yaml` to inherit its locale config from the podspec (and hence docs in that collection which don’t specify their own locales should use the podspec locales too).

However this doesn’t happen; at a minimum, an empty `localization` definition is required:

```yaml
# mycollection/_blueprint.yaml
localization: {}
```

From a quick try, it looks like this would solve it — but could do with some tests:

```diff
diff --git a/grow/pods/collection.py b/grow/pods/collection.py
index c5b3616..5db1321 100644
--- a/grow/pods/collection.py
+++ b/grow/pods/collection.py
@@ -277,24 +277,18 @@ class Collection(object):

     @utils.cached_property
     def locales(self):
         if self.localization:
             if self.localization.get('use_podspec_locales'):
                 return self.pod.list_locales()
             try:
                 return locales.Locale.parse_codes(self.localization['locales'])
             except KeyError:
-                # Locales inherited from podspec.
-                podspec = self.pod.get_podspec()
-                config = podspec.get_config()
-                if ('localization' in config
-                        and 'locales' in config['localization']):
-                    identifiers = config['localization']['locales']
-                    return locales.Locale.parse_codes(identifiers)
-                raise NoLocalesError('{} has no locales.')
-        return []
+                pass
+        # Locales implicitly inherited from podspec.
+        return self.pod.list_locales()
```

jeremydw:
Simplified the fallback structure, and updated with tests to cover several different explicit cases, as covered by https://github.com/grow/grow/pull/267

As mentioned, please take a look and let's reopen this if the tests and behavior are not sufficient. Thanks!

Issue #214, opened 2016-05-08
Allowing documents to declare themselves as having _no_ locales


stucox:
Suggested on #206, to facilitate locale-specific content without overriding (composition rather than inheritance), e.g.:

```yaml
---
$locale: null
foo: base
---
$locale: en
foo: en_foo
---
$locale: fr
foo: fr_foo
---
```

---

My 2 cents: I don’t think this is necessary.

It’s rare you’ll have a document with completely different content in different locales*.

* When content _IS_ completely different, one could argue these are actually different documents, so should be separate files.
* When there is some overlap, it makes sense to have the overlapping content in the base document and mark that document as having the locales for which the content overlaps.

_* assuming documents are always authored in the same language, e.g. US English, and translation is handled entirely by PO files — which is the pattern followed in my projects._

So I don’t think there’s a good real-world use-case.

---

Related question:

**Should a document part be able to declare a locale which _isn’t_ declared in the base document?**

* If we allow `$locale: null`, the answer to this has to be _yes_.
* Allowing this is more consistent with the fact that a collection’s blueprint can declare locales which aren’t declared in podspec, a document can declare locales not declared in the blueprint, etc
* However, this might make documents less readable: it becomes less obvious which locales this page will be built for (5 locales might be declared at the top, but another declared in a later part).
* Allowing this might also complicate Grow’s internal implementation.
* If Grow _disallowed_ this, it could catch some authoring mistakes.

Maybe this is another case where projects should be able to choose which pattern they prefer, either with a setting, or with a custom test?

jeremydw:
With #267 submitted, documents and collections can now disable localization (e.g. declare no locales) with the following syntax:

```
$localization:
  locales: []
```

If `$localization: locales` is omitted, the document or collection will fall back to its parent. In the case of a document, the collection's `$localization` will be used, and in the case of a collection, the podspec's `$localization` will be used.

Note there is not presently any defined `$localization` behavior for subcollections, so the document -> collection -> podspec fallback behavior is adhered to universally.

For content overrides, one can override content in the default locale in the following manner:

```
foo: bar
foo@en_US: bar for en_US
foo@de: bar for de
```

```
# en_US
{{doc.foo}} -> "bar for en_US"

# de
{{doc.foo}} -> "bar for de"

# fr
{{doc.foo}} -> "bar"
```

[...]

Issue #180, opened 2016-02-23
Allow overriding of default_locale content


WillsB3:
Say I have this document:

```yaml
$localization:
  default_locale: en_us
  locales: 
    - en_gb
    - en_us
    - fr_fr

sections:
  header:
    text@: "Some header text"
  promo:
    heading@: "Heading"
```

The following does not work (and I kind of get why, since the default_locale document is the "base" document, and you can't specify a locale override more than once...) but it feels like it should:

```yaml
$localization:
  default_locale: en_us
  locales: 
    - en_gb
    - en_us
    - fr_fr

sections:
  header:
    text@: "Some header text"
  promo:
    heading@: "Heading"

---
$locale: en_us
  something_not_in_the_base_document: "Hi!"
```

Currently, the override content for `en_us` above is ignored, because the locale I'm trying to override is the same as the default locale. I feel like the "default" document should be more of an abstract thing, and not be tied so specifically to the default locale (if that makes sense).

Wondering if this impacts/is impacted by #142...

jeremydw:
I worked on this a bit with @meizon a few months ago. After going back and forth (and exactly agreeing with your reasoning as why you conceptually *shouldn't* be able to do this), we decided to add support for it because the feature would be so useful.

So, I actually think we added support for this in https://github.com/grow/pygrow/releases/tag/0.0.56.

The catch is that you must specify the `$localization: default_locale` explicitly in the document's top frontmatter (just like in your example).

This testdata (https://github.com/grow/pygrow/blob/master/grow/testing/testdata/pod/content/localized/localized.yaml) confirms the ability to override the default locale, so this feature should exist already. (And, I think we need to add some explicit documentation about it to grow.io).

If it's not working for you, can you post a sample YAML and we'll take a look? Thanks!

jeremydw:
Just pinging on this question @WillsB3 -- I actually met with Will T and dcgauld last week and we came up with a reasonable workaround for https://github.com/grow/pygrow/issues/142 -- so I'll be updating that bug this week with the plan.

WillsB3:
That's great news. Thanks for the update.

jeremydw:
This has been implemented, and folks will be able to override default content on a case-by-case basis with our new yaml syntax, e.g.:

```
foo: bar
foo@en_US: bar for en_US
```

PR #267, opened 2016-08-03
Simplify and test localization fallback

### LINKED ISSUE #157: Grow extract extracts strings for non-specified locales

jeremydw:
Simplify localization fallback between documents, collections, podspec; remove validation as it does not account for sub-collections without blueprints, add extensive test coverage for all the ways one could localize (or disable localization of) documents.

This PR:
- *Removes* the requirement of bueprints to, at a minimum, specify `$localization: {}` in order to opt-in to localization for the entire collection, even if the podspec is localized. This now requires blueprints to set `$localization: locales: []` to disable localization, if it was enabled at the podspec-level. #265 
- Permits documents to be localized, even if blueprints or the podspec aren't, by specifying `$localization:locales` at the document-level. #204 
- Allows one to disable localization for a document by using `$localization: locales: ~`. #214 
- Enables one to override the value of a key in the default locale using `<key>@<default locale>: <value>` without modifying the value of keys in other locales. #180 
- Removes locale validation since I discovered too many cases of sub-folders in collections being used without blueprints in those folders, which is basically undefined behavior. I don't want to break the use of this, so I removed the locale validation. 
- Should add enough test coverage to close #157 

At a high level (this will be documented after this PR is merged), here's how document localization works. This is not really different from practical use cases before, with the exception of defaulting to inheriting the podspec's locales if `$localization` isn't present in a collection but is in the podspec.

1. Uses a document's `$localization:locales` if present.
1. If not present, falls back to the blueprint's `$localization:locales`.
1. If blueprint's `$localization:locales` is not present, use locales from the podspec.

The use of `$localization:locales` -> `[]` or `nil` localization disables localization.

@stucox please take a look when you get a chance, particularly at all the tests in `documents_test.py` since they describe and verify the above notes. Thanks!

### DIFF
--- a/grow/pods/collection.py
+++ b/grow/pods/collection.py
@@ -287,13 +287,10 @@ class Collection(object):
 
     @utils.cached_property
     def locales(self):
-        # Require $localization in blueprint for localization.
-        if self.localization is not None:
-            if 'locales' in self.localization:
-                identifiers = self.localization['locales'] or []
-                return locales.Locale.parse_codes(identifiers)
-            return self.pod.list_locales()
-        return []
+        if self.localization and 'locales' in self.localization:
+            codes = self.localization['locales'] or []
+            return locales.Locale.parse_codes(codes)
+        return self.pod.list_locales()
 
     def to_message(self):
         message = messages.CollectionMessage()
--- a/grow/pods/collection.py
+++ b/grow/pods/collection.py
@@ -287,20 +287,12 @@ class Collection(object):
 
     @utils.cached_property
     def locales(self):
-        if self.localization:
-            if self.localization.get('use_podspec_locales'):
-                return self.pod.list_locales()
-            try:
-                return locales.Locale.parse_codes(self.localization['locales'])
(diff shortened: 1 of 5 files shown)


jeremydw:
@stucox and @WillsB3 I have some time today to dedicate to this so I'm going to go ahead and merge this in and close the associated issues I tagged in this bug so that this has some soak time in `master` before we end up making a release (which we'll do hopefully shortly after you take a look at this PR).

When you get a chance, please take a look at the commits, especially the test coverage I added to cover several different `$localization` cases, and we can make post-merge fixes/enhancements if you identify anything. Thanks a ton!

PR #287, opened 2016-08-28
Optimization

kezlya:
First draft of using global dictionary for parsed, localized yamls:
Performance improved by ~ 30%

===== Before optimization =====
Building: 1419/1419 (in 0:00:07)
Building: 16873/16873 (in 0:01:24)

===== After optimization ======
Building: 1419/1419 (in 0:00:03)
Building: 16873/16873 (in 0:00:52)

before

after

### DIFF
--- a/grow/pods/documents.py
+++ b/grow/pods/documents.py
@@ -108,6 +108,18 @@ class Document(object):
     def format(self):
         return formats.Format.get(self)
 
+    @utils.cached_property
+    def virtual_key(self):
+        last_mod = 0
+        try:
+            last_mod = self.pod.file_modified(self.pod_path)
+        except OSError:
+            pass
+        return "{0}|{1}|{2}"\
+            .format(self.root_pod_path,
+                    self._locale_kwarg,
+                    last_mod)
+
     @property
     def url(self):
         path = self.get_serving_path()
--- a/grow/pods/formats.py
+++ b/grow/pods/formats.py
@@ -17,6 +17,7 @@ import yaml
 
 BOUNDARY_REGEX = re.compile(r'^-{3,}$', re.MULTILINE)
 PATH_LOCALE_REGEX = re.compile('(.*)@([^\.]*)\.(.*)')
+GLOB_YAMLS_DIC = {}
 
 
 class Error(Exception):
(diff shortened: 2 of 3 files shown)


jeremydw:
This is great. The benchmark is also great. I was just thinking about doing an optimization almost exactly like this for "cached" environments – e.g. when you know that the files won't be changing beneath you.

Two things:

1) Can you verify this change works as intended in the dev server? In other words, if I change yaml and refresh the page, do you see the updated content? (My thought is no, because the cache key is the pod path, and I don't think updates to file content will invalidate the cache.)

2) Rather than using a global you can use the `@utils.memoize` decorator on a function. The `@utils.memoize` decorator uses the function arguments as the cache key, and it also supports a `.reset()` method that you can use to clear the cache if necessary.

My thought: Verify (1). If it's no good, then at a minimum we can speed up `grow build` by enabling this for cached environments. You can determine if an environment is cached by checking `pod.env.cached`.

LMK what you think. Thanks!!

kezlya:
I'm still doing some clean up, refactoring and fixing tests, should work even better. I created this PR just to start a conversation and verify with contributors that this changes going to the right direction.

1)  I'll add "file modified time" to the "key" and that should do a trick for the dev server.
2) I had same thoughts and started from using `@utils.memoize` but `pod` parameter is always different so it always create new key. In my upcoming commits I'm back to use `@utils.memoize` but with custom key, so we don't have to deal with global variables as you mention.

kezlya:
1) Done, tested on dev server
2) Done, removed global variable, using dictionary in Pod
All tests pass :)
Functionality extended for all files formats (.html, .md, etc) not just Yaml

jeremydw:
This is a great optimization — thanks for making it. I'm going to test it out a bit more tomorrow and then merge it on in. I'll rebase the work I'm doing in `feature/cache` against this and make sure to combine all the improvements!

jeremydw:
Tested this out a bit and audited the code, I think this should be all good to merge. Let me know if you're planning to add more to this PR and I'll leave it open otherwise I'll merge it on in. Thanks!

kezlya:
Yeap, this is good to go from my side. Ready to merge :)

PR #299, opened 2016-09-21
Implement ability to specify extract settings in podspec.

jeremydw:


stucox:
Nice! LGTM

jeremydw:
Thanks – this is an important feature in order to implement continuous translation!

Issue #289, opened 2016-08-30
`grow extract --localized` doesn't behave correctly for subcollections


stucox:
Looks like `grow extract --localized` extracts terms in docs in subcollections for all podspec locales, rather than the locales declared by the nearest blueprint (which is what I’d expect).

I’m afraid I haven’t had time to create a test case yet, but I have a structure similar to:

```
/
├─ content/
│   ├─ mycollection/
│   │   └─ mysubcollection/
│   │       ├─ _blueprint.yaml
│   │       └─ mydoc.yaml
│   :
├─ podspec.yaml
:
```

where:

* `mydoc.yaml` doesn’t declare a `$localization` config

* `_blueprint.yaml` declares:

    ```yaml
    localization:
      default_locale: en_US
      locales:
        - de_DE
        - en_US
    ```

* podspec.yaml declares

    ```yaml
    localization:
      default_locale: en_US
      locales:
        - de_DE
        - en_US
        - fr_FR
    ```

Tagged strings in `mydoc.yaml` are unexpectedly extracted for `fr_FR`.

stucox:
Also, it seems declaring a `path` property in the localization config in both `mycollection` and `mysubcollection` causes _two_ copies of the page to be built (one at each path).

jeremydw:
Your follow-up comment provides the best clue here. I suspect this behavior should be fixed with https://github.com/grow/grow/pull/303. In that PR, we do just what you expect – we find the closest `_blueprint.yaml` and use that for the document.

Previously, pages in subcollections that had their own `_blueprint.yaml` were being built twice, which also means they were being extracted twice -- once for locales defined in the top-level `_blueprint.yaml` and once for locales defined in the nearer one.

jeremydw:
I just committed https://github.com/grow/grow/pull/303 -- so I'm going to go ahead and close this out as I want to release a new version of Grow with many improvements today/this weekend. When you get a chance, please try this out on HEAD (or in the release), and let me know if it fixes things up with you. :) If the fix wasn't sufficient, go ahead and reopen this issue.

stucox:
Excellent 👍  Thanks @jeremydw!

PR #303, opened 2016-09-23
Improved path-clobbering validation

### LINKED ISSUE #278: Locale clobbering when no localized path is specified
With the changes made in https://github.com/grow/grow/issues/204 we now have an issue where content that is localized but no localized path is specified will be clobbered by the last locale in the document's list of locales. If no localized path is specified, when a page is being rendered, the default locale should always be used.

jeremydw:
Fixes #278


jeremydw:
I suspect this should also fix #289.

PR #311, opened 2016-10-10
Updating the fuzzy matching argument for the extract command

Zoramite:
When calling the extract command the argument value for fuzzing matching is being used and ignoring the extract configuration.

Decided to also change the name of the argument to be more consistent to the rest of the variables.


jeremydw:
Looks good. 👍

jeremydw:
(Feel free to merge.)

PR #312, opened 2016-10-10
Safer translation

Zoramite:
Updating the `upload_translations` command to be able to `extract` the catalogs before uploading.

Also set the default for the `download` and `extract` commands to be ON to help prevent accidental translation loss.


jeremydw:
👍

PR #316, opened 2016-10-12
Adding a column to the spreadsheet for the location of the message.

Zoramite:
Also added an option to include the obsolete messages as part of the upload.

### DIFF
--- a/grow/commands/upload_translations.py
+++ b/grow/commands/upload_translations.py
@@ -23,8 +23,13 @@ import os
                    ' to uploading.')
 @click.option('--extract', '-x', default=True, is_flag=True,
               help='Whether to extract translations prior to uploading.')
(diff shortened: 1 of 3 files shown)


jeremydw:
Excellent -- tested this myself and it works well.

I want to ensure the editing experience is good for translators, and that the column isn't too distracting from performing the translation work.

While you've got these files open, can you make the cell background `RGBA(243, 243, 243)` (this matches the Google Sheets grey color) and can you also set the column width to 250?

Zoramite:
I've Updated the location width to be smaller. For the location, it will be less distracting to have the text color a grey color, so I did that instead. Then we can also use the background colors to highlight the obsolete translations easier.

jeremydw:
👍 

Tried this out. Looks good! Thank you. Feel free to merge it on in.

PR #320, opened 2016-10-14
Add support for regexes in locale-override yaml syntax.

### LINKED ISSUE #319: FR: Wildcard localized keys
I find myself doing this a lot when EN source strings change, but translations aren't ready yet:

```
foo@: Old text
foo@en: New text
foo@en_CA: New text
foo@en_GB: New text
...
```

Any chance something like this might be possible or useful?

```
foo@: Old text
foo@en: New text
foo@en_*: New text
```

jeremydw:
Fixes #319 

@stevenle @Zoramite – can you look at this syntax? In practice this seems to work really well and be a good, flexible solution I'm just worried that the following YAML syntax is too verbose. Any thoughts or is it OK?

```
foo@: Base foo.
foo@/de/: Foo for any locale that matches regex `de`.
foo@/it|fr|ca/: Foo for any locale that matches regex `it|fr|ca`.
```

### DIFF
--- a/grow/common/utils.py
+++ b/grow/common/utils.py
@@ -320,13 +320,8 @@ def untag_fields(fields, locale=None):
             updated_localized_paths.add((path, key))
             return key, value
         untagged_key, locale_from_key = match.groups()
-        if (locale
-                and locale_from_key.startswith('/')
-                and locale_from_key.endswith('/')):
-            locale_regex = locale_from_key[1:-1]
-            if not re.match(locale_regex, locale):
-                return False
-        elif locale_from_key != locale:
+        locale_regex = r'^{}$'.format(locale_from_key)
+        if not locale or not re.match(locale_regex, locale):
             return False
         updated_localized_paths.add((path, untagged_key.rstrip('@')))
         return untagged_key, value
--- a/grow/common/utils.py
+++ b/grow/common/utils.py
@@ -34,7 +34,7 @@ except ImportError:
     from yaml import Loader as yaml_Loader
 
 
-LOCALIZED_KEY_REGEX = re.compile('(.*)@([\w|-]+)$')
+LOCALIZED_KEY_REGEX = re.compile('(.*)@([^@]+)$')
 SENTINEL = object()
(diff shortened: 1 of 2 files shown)


stevenle:
How about just having it accept regexes by default? These seem fine to me:

```yaml
foo@en.*: bar
foo@it|fr: bar
```

jeremydw:
I thought about that, it's definitely a lot cleaner, but then `foo@fr` would match `foo@fr_CA` and `foo@fr` by default. Technically that change might cause backwards incompatibility -- maybe it's OK?

stevenle:
Is there a python regex function that forces the regex to match the entire string, so that `fr` doesn't match `fr_CA`?

Zoramite:
Are you trying to easily handle wildcards? Why not modify the regex a little to match it as a separate group at the end that can easily be checked to see if it is a wildcard?

ex: `(.*)@([\w|]+)(_\*)?$`

Checking the third group for a match would let you know that you are looking at a `_*` ending and that you should match all of them that match the language(s) in the second group?

Of if you want to go full regex, it would make the syntax cleaner and you can use this instead: 

    locale_regex = '^%s$' % locale_from_key

And it should be backwards compatible.

Then you would have to specifically wildcard it to get it to match more than a straight string. Ex: `foo@fr` would not match `fr_CA`. But if you did `foo@fr.*` or `foo@fr_.*` it would match all of the `fr` variants. Or even more complex: `foo@(fr|it|es).*` would be valid as well. Or match all CA versions... for whatever reason `foo@.*_CA`.

stevenle:
FWIW, I'm personally in favor of the `locale_regex = '^{}$'.format(locale_from_key)` soln. Makes sense to me that `foo@fr` doesn't match `fr_CA`.

jeremydw:
@Zoramite's suggestion looks good. I think common use cases would be like the following:

```
foo@fr
foo@fr_.*
foo@.*_CA
foo@(.*_CA|.*_DE)
```

So as long as we support the ones like this we should be OK. @Zoramite will you take over the code changes from here?

Zoramite:
@jeremydw I've updated it to use the specific regex and added a couple of more tests to make sure that it handles the straight text match along with the regex wildcards.

PR #322, opened 2016-10-14
Updated the upload command to correctly store the extracted messages.

Zoramite:
The catalogs were not being correctly updated with the latest message that were extracted when using the flag.

Also, added the `--no-` version of the extract and download commands.


jeremydw:
👍

PR #323, opened 2016-10-14
Updating the way that translation spreadsheets are updated.

Zoramite:
Previously all translation information was blindly written over the existing data. This caused a couple of issues such as when several translations are added past the buffer cells or the comments in a spreadsheet keeping the same position but having a different cell value hence losing the context of the comment.

The updates are focused on being a bit more selective on how it updates the values already present, adds new values, and removes obsolete messages. By being purposeful with the order of operations the comments or other cell specific meta information is preserved.

### DIFF
--- a/grow/commands/upload_translations.py
+++ b/grow/commands/upload_translations.py
@@ -24,13 +24,13 @@ import os
                    ' to uploading.')
 @click.option('--extract/--no-extract', '-x', default=True,
               help='Whether to extract translations prior to uploading.')
-@click.option('--prune-obsolete', default=False, is_flag=True,
+@click.option('--prune', default=False, is_flag=True,
               help='Whether to remove obsolete messages from spreadsheet.'
                    ' Normally message may be removed and readded periodically'
                    ' so they are not removed, but can be removed to clean up'
                    ' translations but cannot be retrieved once pruned.')
 def upload_translations(pod_path, locale, force, service, update_acl,
-                        download, extract, prune_obsolete):
+                        download, extract, prune):
     """Uploads translations to a translation service."""
     root = os.path.abspath(os.path.join(os.getcwd(), pod_path))
     pod = pods.Pod(root, storage=storage.FileStorage)
@@ -54,4 +54,4 @@ def upload_translations(pod_path, locale, force, service, update_acl,
     else:
         # TODO Remove the download from the upload arguments to clean it up.
         translator.upload(locales=locale, force=force, verbose=True,
-                          download=download, prune_obsolete=prune_obsolete)
+                          download=download, prune=prune)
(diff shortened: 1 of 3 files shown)


grow/translators/google_sheets.py
```diff
@@ -292,39 +296,147 @@ def _create_sheets_request(self, catalogs, source_lang, spreadsheet_id,
...
+                requests.append({
+                    'appendCells': {
+                        'sheetId': sheet_id,
+                        'fields': '*',
+                        'rows': row_data,
+                    },
+                })
+
+            # Remove obsolete rows if not included.
+            if not include_obsolete and len(removed_rows):
```
Zoramite: Alright, I've updated it to make it a `prune_obsolete` and removed the `include_obsolete` from the upload command. It will now just use the config version of the setting for doing the extraction.

Also made it so that the option to prune is off by default.

grow/commands/upload_translations.py
```diff
@@ -23,21 +23,21 @@
                    ' to uploading.')
 @click.option('--extract', '-x', default=True, is_flag=True,
               help='Whether to extract translations prior to uploading.')
-@click.option('--include-obsolete/--no-include-obsolete', default=None,
-            is_flag=True,
-            help='Whether to include obsolete messages. If false, obsolete'
-                 ' messages will be removed from the upload. By'
-                 ' default, Grow cleans obsolete messages from the upload')
+@click.option('--prune-obsolete', default=False, is_flag=True,
```
Zoramite: I was trying to be more explicit on what it does, but we can go with `prune`

jeremydw: Thanks -- when it comes to the command-line I've found that "less is more" and it's easier to explain when the flags are short and sweet. :)

Zoramite:
How does the rest of the commit look? Any other changes you would suggest?

grow/translators/google_sheets.py
```diff
@@ -54,13 +55,12 @@ def _download_sheet(self, spreadsheet_id, locale):
...
                     'Translation for {} not found.'.format(locale))
             raise
-
         return resp['values']
 
     def _download_content(self, stat):
         spreadsheet_id = stat.ident
         values = self._download_sheet(spreadsheet_id, stat.lang)
-        source_lang, lang = values.pop(0)
+        source_lang, lang, _ = values.pop(0)
```
jeremydw: Just a note, no action needed – I wonder if we should make this accept variable numbers of columns to increase the robustness. Again, no action needed but this would require updating if future enhancements make changes to the columns.

jeremydw:
Just gave it a good look-over. All looks good, thanks for keeping the existing rows in place (to support user-added comments!) and the addition of new strings at the end. :)

👍

grow/translators/google_sheets.py
Zoramite: From what I can tell it does fine with more columns since we have to specifically request which columns we want (ex: A:C in this case). So additional columns should not matter, as long as the first three are ours.

PR #324, opened 2016-10-17
Adding a custom error for when translation sources are not found.

### LINKED ISSUE #314: Clean up error messages when using `grow download_translations`
When `grow download_translations` is run without specifying `--locale`, Grow attempts to download translations for all locales. Sometimes, a Google Sheet might not have a tab for each locale. This is OK and supported behavior -- we just need to clean up the output when this scenario is encountered.

Sample output today:

```
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/Users/jeremydw/git/grow/bin/../grow/common/utils.py", line 392, in run
    super(ProgressBarThread, self).run()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/Users/jeremydw/git/grow/bin/../grow/translators/base.py", line 100, in _do_download
    new_stat, content = self._download_content(stat)
  File "/Users/jeremydw/git/grow/bin/../grow/translators/google_sheets.py", line 53, in _download_content
    values = self._download_sheet(spreadsheet_id, stat.lang)
  File "/Users/jeremydw/git/grow/bin/../grow/translators/google_sheets.py", line 48, in _download_sheet
    spreadsheetId=spreadsheet_id, range=rangeName).execute()
  File "/Users/jeremydw/git/grow/env/lib/python2.7/site-packages/oauth2client/util.py", line 135, in positional_wrapper
    return wrapped(*args, **kwargs)
  File "/Users/jeremydw/git/grow/env/lib/python2.7/site-packages/googleapiclient/http.py", line 723, in execute
    raise HttpError(resp, content, uri=self.uri)
HttpError: <HttpError 400 when requesting https://sheets.googleapis.com/v4/spreadsheets/xxx/values/%27en_US%27%21A%3AB?alt=json returned "Unable to parse range: 'en_US'!A:B">
```

Instead of returning this error message, we should catch `"Unable to parse range"` and use instead:

```
self.pod.logger.info('No worksheet to download: en_US')
```

Then, sample output will be like the following (instead of including a traceback:

```
[13:39:31] No worksheet to download: en_US
[13:39:31] Imported 520 translations: de_DE
[13:39:31] Imported 525 translations: en_GB
[13:39:32] Imported 521 translations: fr_CA
[13:39:32] Imported 524 translations: en_AU
```

Zoramite:
Using for when a spreadsheet type of translation does not have a sheet for the locale being requested or the sheet has been removed.

Closes #314

### DIFF
--- a/grow/translators/base.py
+++ b/grow/translators/base.py
@@ -2,6 +2,7 @@ from protorpc import message_types
 from protorpc import messages
 from protorpc import protojson
 from grow.common import utils
+from grow.translators import errors as translator_errors
 import copy
 import json
 import logging
@@ -97,7 +98,13 @@ class Translator(object):
         langs_to_translations = {}
         new_stats = []
         def _do_download(lang, stat):
-            new_stat, content = self._download_content(stat)
+            try:
+                new_stat, content = self._download_content(stat)
+            except translator_errors.NotFoundError:
+                text = 'No translation source for {} locale.'
+                self.pod.logger.info(text.format(lang))
+                return
+
             new_stat.uploaded = stat.uploaded  # Preserve uploaded field.
             langs_to_translations[lang] = content
             new_stats.append(new_stat)
--- a/grow/translators/base.py
+++ b/grow/translators/base.py
@@ -101,7 +101,7 @@ class Translator(object):
             try:
                 new_stat, content = self._download_content(stat)
             except translator_errors.NotFoundError:
-                text = 'No translation source for {} locale.'
+                text = 'No translations to download for: {}'
                 self.pod.logger.info(text.format(lang))
                 return
 
--- a/grow/translators/errors.py
+++ b/grow/translators/errors.py
@@ -1,0 +1,6 @@
+class Error(Exception):
+    pass
+
+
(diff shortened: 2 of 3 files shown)


grow/translators/base.py
```diff
@@ -97,7 +98,13 @@ def download(self, locales, save_stats=True, inject=False):
         langs_to_translations = {}
         new_stats = []
         def _do_download(lang, stat):
-            new_stat, content = self._download_content(stat)
+            try:
+                new_stat, content = self._download_content(stat)
+            except translator_errors.NotFoundError:
+                text = 'No translation source for {} locale.'
```
jeremydw: Just a slight nit on the error message -- this should probably be something like `No translations to download for: {}`. Because the term "source" might be overloaded and confused with the source language, etc.

Zoramite: K, I've updated this message.

jeremydw:
👍

PR #325, opened 2016-10-17
Adding the ability to use `none` or no language for the code block.

### LINKED ISSUE #307: "sourcecode:none" error only flagged in Mac OS
When building the [AMP docs](https://github.com/ampproject/docs) on Mac OS, I receive the following error:

grow.pods.errors.BuildError: Error building <Rendered(view='/views/doc.html', doc='/content/docs/reference/extended/amp-user-notification.md')>: 'grow.pods.documents.Document object' has no attribute 'html'

The reason for the error is that the code block has the lang specified as "none".  For example:

[sourcecode:none]
https://foo.com/api/show-api?timestamp=1234567890&elementId=notification1&ampUserId=cid-value
[/sourcecode]

The error is correct and flagged the item correctly; however, when I build the same docs on Ubuntu or through the [Travis production job] (https://travis-ci.org/ampproject/docs/builds/163559567), no error occurs.

Grow version: 0.66
Mac OSX v: 10.11.6

FYI: I have since fixed the code block to specify the language as "html".

### LINKED ISSUE #321: Allow for unspecified languages in codeblocks
In most of our codeblocks, we can specify a language but in some, we just want a plain non-syntax highlighted block, for example, showing a URL in a codeblock.  What option or setting can be enabled to allow for plain codeblocks?

Example:

```
[sourcecode]
# code block
[/sourcecode]
```

Zoramite:
Instead of failing it should now use a default of the 'text' formatting which does not have any coloring.

This makes `[sourcecode:none]` and `[sourcecode]` both valid.

Closes #307
Closes #321

### DIFF
--- a/grow/common/markdown_extensions.py
+++ b/grow/common/markdown_extensions.py
@@ -128,7 +128,7 @@ class CodeBlockPreprocessor(preprocessors.Preprocessor):
         src/e79a7126551c39d5f8c1b83a79c14e86992155a4/external/markdown-processor.py
     """
     KIND = 'sourcecode'
-    pattern_tag = re.compile(r'\[sourcecode:(.+?)\](.+?)\[/sourcecode\]', re.S)
+    pattern_tag = re.compile(r'\[sourcecode(:.*?)\](.+?)\[/sourcecode\]', re.S)
     pattern_ticks = re.compile(r'```(.+?)\n(.+?)\n```', re.S)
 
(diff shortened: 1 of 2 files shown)


jeremydw:
👍 Thanks! Merge it on in.

bpaduch:
Does this apply to the github markdown too?  For example using backticks only with no code specified?

Issue #317, opened 2016-10-12
Meta Information on the Translation Spreadsheets


Zoramite:
Currently the translation export to google sheets just does straight data. To improve this we can add conditional formatting to the spreadsheet to highlight specific areas.
-  **Missing Translations:** Highlighting the missing translations makes it easier to pick out the work left to do.
-   **Obsolete Messages:** Highlight the messages that are no longer used in any templates.

PR #326, opened 2016-10-18
Separate the formatting of sheets from the data.

Zoramite:
Also revamping the way that batch requests are created to make the number of server requests smaller.

The creation of new language sheets is also improved to be created as part of the same batch request as other updates making the creation and update process atomic.

Currently updating the meta (such as styling) is available under the `--update_meta` flag. In the future want to make this process smarter to automatically apply any time a new version of grow is used to upload translations.

Part of #317


jeremydw:
👍

PR #330, opened 2016-10-20
Fixing having too few columns for uploading translations to Google Sheets

Zoramite:
Older version of the spreadsheet may be missing a column (ex: locations) or using existing spreadsheets may only have the source and translation columns. These changes should enable the upload command to detect when there is missing columns and add them to the sheet and fill in the missing column header.

### DIFF
--- a/grow/translators/google_sheets.py
+++ b/grow/translators/google_sheets.py
@@ -47,6 +47,7 @@ class GoogleSheetsTranslator(base.Translator):
         'green': .6196,
     }
     KIND = 'google_sheets'
+    COLUMN_COUNT = 3
     HEADER_ROW_COUNT = 1
     has_immutable_translation_resources = False
     has_multiple_langs_in_one_resource = True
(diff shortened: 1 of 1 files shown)


grow/translators/google_sheets.py
```diff
@@ -66,6 +67,13 @@ def _download_sheet(self, spreadsheet_id, locale):
                 raise translator_errors.NotFoundError(
                     'Translation for {} not found.'.format(locale))
             raise
+
+        # Check for spreadsheets that are missing columns.
+        column_count = len(self.HEADER_LABELS)
+        if len(resp['values'][0]) < column_count:
+            missing_columns = [None] * (column_count - len(resp['values'][0]))
+            resp['values'][:]=[i + missing_columns for i in resp['values']]
```
jeremydw: Style nit on the spacing between `=`. Other than that 👍 and merge it on in.

grow/translators/google_sheets.py
```diff
@@ -48,6 +48,7 @@ class GoogleSheetsTranslator(base.Translator):
     }
     KIND = 'google_sheets'
     HEADER_ROW_COUNT = 1
+    HEADER_LABELS = ['', '', 'Location']
```
jeremydw: Maybe add a comment here for "# Source locale, translation locale, message location."

PR #331, opened 2016-10-20
Removing the download option as an argument to the upload method.

Zoramite:
Instead moving the logic to download into the `upload_translations` command.


jeremydw:
👍

PR #333, opened 2016-10-21
Updating the update check to verify the platform is available.

Zoramite:
In case there is a problem with a release process not uploading the correct binary, the auto-update check should only show when the release has the correct binary asset for the current system.


jeremydw:
👍

PR #337, opened 2016-10-27
UI Menu Changes

Zoramite:
Updating the grow menu that is used used on dev builds.

Currently it supports the edit and translation functionality, in the future it will expand as we have other tools for development (ex: analytics highlighting and image testing).

In the future we want to be able to add this menu to the staging builds for some of the tools (ex: analytic highlighting and image testing).

### DIFF
--- a/grow/ui/js/ui.js
+++ b/grow/ui/js/ui.js
@@ -3,32 +3,53 @@ grow.ui = grow.ui || {};
 window.grow = grow;
 
 grow.ui.main = function(settings) {
-  if (!settings.injectEditUrl && !settings.injectTranslateUrl) {
-    return;
-  }
+  var buttonEl;
   var buttonsEl = document.createElement('div');
-  buttonsEl.setAttribute('class', 'grow__buttons');
+  buttonsEl.setAttribute('class', 'buttons');
+  buttonsEl.addEventListener('mouseover', function() {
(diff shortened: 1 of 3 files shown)


jeremydw:
Initial work is looking good, three initial pieces of feedback:

image

(1) I think we should probably make the Grow logo smaller.
(2) The "X" overlaps the logo.
(3) When the pencil returns back into the logo, it doesn't slide quite down far enough.

I wonder if we should be pushing the logo itself there, or if that might be logo overage. I don't think average site editors would know what that is -- maybe we use an "edit" button (like how Material Design uses a "+") and that expands the options from there? I'd need to think about that some more.

Zoramite:
1. The Grow logo is now smaller
1. I've updated the close icon to close pull out to the left, per our offline discussion.
1. Fixed the formula that is used to calculate where the icons should be based from. This should be fixed for realz this time.

For now I'd like to stick with the logo. Originally I was going to do something like have it become the some other menu, but there were a couple of issues with that: 1) mobile would struggle since there is no :hover so it acts as a way to trigger on mobile. 2) The menu items can change based on the context, so having it morph into another button makes it tricky across multiple projects. 3) The grow is awesome, promote it :)

If we really don't like using the logo we can switch it out with some generic menu icon, like the three dots...

PR #338, opened 2016-11-02
Grow UI Tool for Google Analytics Autotrack

Zoramite:
Adding a sample GA Inspection tool that highlights and shows basic Google Analytics data attributes when hovering.

### DIFF
--- a/grow/ui/sass/tool_ga_inspect.sass
+++ b/grow/ui/sass/tool_ga_inspect.sass
@@ -1,0 +1,19 @@
+.grow__tool__ga
+  &--active
+    [data-g-event]
+      background: repeating-linear-gradient(-45deg, #FBE9E7, #FFCCBC 5px, white 5px, white 10px)
+
+      &:hover::after
+        background: #E64A19
+        border-radius: .25em
+        bottom: 50px
+        color: #ffffff
+        content: '[Event] ' attr(data-g-event) ' [Action] ' attr(data-g-action) ' [Label] ' attr(data-g-label)
+        font-size: 1.3rem
+        font-weight: bold
+        left: 50px
+        padding: 1em
+        position: fixed
+
+.grow__icon_tool_ga_inspect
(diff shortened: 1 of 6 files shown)


grow/ui/sass/tool_ga_inspect.sass
```diff
@@ -0,0 +1,19 @@
...
+  &--active
+    [data-g-event]
+      background: repeating-linear-gradient(-45deg, #FBE9E7, #FFCCBC 5px, white 5px, white 10px)
+
+      &:hover::after
+        background: #E64A19
+        border-radius: .25em
+        bottom: 50px
+        color: #ffffff
+        content: '[Event] ' attr(data-g-event) ' [Action] ' attr(data-g-action) ' [Label] ' attr(data-g-label)
```
jeremydw: Hm -- I don't think we should add this to grow itself, because the `data-g-event` syntax is specific to `autotrack.js` -- and it's not even a part of airkit (yet).

Unless we want to be presumptuous about how people tag their sites, I think this needs to be configurable somehow to avoid making assumptions. That's one reason I thought it might be a good reason to make this part of airkit rather than grow, or at least start by adding support for `data-g-*` in airkit first, and then more strongly supporting the airkit+grow combination. Thoughts?

Zoramite: I have been thinking this morning that we could make the grow tools being used part of the configuration (podspec) then have them be off by default if they are not applicable to all of the grow sites. Then we can be a bit more modular with specific tools like this one and make them optional/extensible.

jeremydw: The analytics tracking or experimentation is one I can think of off the top of my head.

Right now unless a website uses `autotrack.js` (which is from Google, not even related to Grow or open source) the `data-g-event` stuff won't actually do anything. So, my thinking is we add support for some equivalent functionality in airkit, that does something similar to `autotrack.js` and then we add support for it in Grow's UI.

Zoramite: Ok, I'll work on changing this PR to focus on using the settings to control tools. Then later when we work on GA in airkit we can update the `ga-inspect` tool to work with that as well.

Zoramite:
So, I am keeping this as an autotrack tool for now and not doing it as a settings for the tools. I'm going to follow it up later to make it a more generic and configurable analytics tool with #342 for controlling the settings.

Zoramite:
Here is an example of the current tool in use.

screen shot 2016-11-09 at 9 53 59 am

Zoramite:
I've separated out the dynamic adding of tools to #343 and changed this PR to be specific to the autotrack highlighting.

Zoramite:
I actually had a different idea for this, stay tuned.

Issue #342, opened 2016-11-09
UI Tool System


Zoramite:
There is a lot of utilities and additional functionality that can be added for the UI tooling in grow.

Work on a system that allows for separation of the tools and the grow binary. Tools should be able to be updated/managed separately. Also third-party and custom tools should be supported.

PR #344, opened 2016-11-10
Initial functionality for supporting the dynamic ui tools

Zoramite:
Reads the list of tools from the podspec and uses it to include the correct files from the project's `node_modules` directory.

This initial version is very opinionated and targeted towards the usage of npm. Future versions should make this path variable based on configuration in the podspec.

Part of #342

### DIFF
--- a/grow/server/main.py
+++ b/grow/server/main.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import jinja2
+import mimetypes
 import re
 import sys
 import traceback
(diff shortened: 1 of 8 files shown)


grow/server/main.py
```diff
@@ -76,6 +77,15 @@ def serve_pod(pod, request, values):
     return response
 
 
+def serve_tool(pod, request, values):
```
jeremydw: For convention I'd say `serve_ui_tool`.

Zoramite: Fixed.

jeremydw:
Alright -- this is looking good now. Feel free to merge it in. Without some use cases I'm not 100% sure if this is the right convention to follow (for example, jinja extensions are NOT convention based and the user has to specify explicitly where they come from) but I'm thinking it could be OK for tools.

This is good for now and we can revisit later if it's not working out!

PR #350, opened 2017-01-01
Accept order_by and reverse arguments to g.collections() and related functions; and some bug fixes

maznu:
In a blog project I have a tree of nested collections containing years, months, and posts.  I wanted to be able to do this:

```
    <dl>
      {% for year in g.collection('blog').collections(order_by="path") %}
      <dt>{{ year.title }}</dt>
      <dd><ul>
        {% for month in year.collections(order_by="path") %}
          <li><a href="{{ month.path }}">{{ month.month }}</a></li>
        {% endfor %}
      </ul></dd>
      {% endfor %}
    </dl>
```

...and with this commit, now I can.

NB: side-effect is that by default collections are sorted by an "order" property (if defined). Given previous behaviour was undefined (semi-random file-system order?), I don't believe this causes any serious regressions.


maznu:
Phew, took a while to get it passed by all the CI checks.  Sadly I wasn't able to get a working test environment locally - so apologies for the mess of commits.  But it's done and works now.  Hope you're happy to pull it.

Zoramite:
What is the benefit to doing this using the collections instead of using a jinja filter?

maznu:
Other functions in grow's API allow order_by; this felt a consistent way of doing that. Functionally, there's no difference, of course.

grow/common/utils.py
```diff
@@ -279,6 +279,7 @@ def slugify(text, delim=u'-'):
     for word in SLUG_REGEX.split(text.lower()):
         word = word.encode('translit/long')
         if word:
+            word = "".join(SLUG_REGEX.split(word)) # deal with smartquotes
```
Zoramite: Can you remove this from this PR? I'm not sure if that is how we want to solve that issue, and it helps to keep the PR cleaner to not have random other fixes in it.

PR #351, opened 2017-01-11
Inject the UI for any pages that have the ui settings.

Zoramite:
Making the UI Tools available to be part of normal builds and for specific deployments with ability for deployment specific settings.

### DIFF
--- a/grow/pods/pods.py
+++ b/grow/pods/pods.py
@@ -302,8 +302,9 @@ class Pod(object):
             tool_path = '{}{}{}'.format(
                 source_prefix, tool_prefix, tool['kind'])
             for root, dirs, files in self.walk(tool_path):
-                if '.git' in dirs:
-                    dirs.remove('.git')
+                for directory in dirs:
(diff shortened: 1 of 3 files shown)


grow/pods/pods.py
```diff
@@ -273,8 +281,49 @@ def export(self):
...
+            source_path = os.path.join(source_root, path)
+            output_path = os.sep + os.path.join(destination_root, path)
+            output[output_path] = self.storage.read(source_path)
+
+        # Add the files from each of the tools.
+        for tool in self.ui.get('tools', []):
+            tool_path = '{}{}{}'.format(
+                source_prefix, tool_prefix, tool['kind'])
+            for root, dirs, files in self.walk(tool_path):
+                if '.git' in dirs:
```
jeremydw: Skip any hidden files by default?

Zoramite: Updating to ignore any directory starting with a `.`.

PR #357, opened 2017-01-23
Feature/travis release

Zoramite:
Attempting to automatically generate the release binaries on tag.


Zoramite:
Yes, it should do a draft release.

PR #358, opened 2017-01-23
Fix stage without explicit webreview destination.

rspier:
This patch allows staging to webreview without an explicit webreview
destination.  Otherwise, there would be errors on trying to reference
undefined fields or objects.

### DIFF
--- a/grow/deployments/destinations/webreview_destination.py
+++ b/grow/deployments/destinations/webreview_destination.py
@@ -73,8 +73,8 @@ class WebReviewDestination(base.BaseDestination):
         return self._webreview
 
     def _get_subdomain(self):
-        repo = common_utils.get_git_repo(self.pod.root)
         if self.config.subdomain_prefix and not self.config.subdomain:
+            repo = common_utils.get_git_repo(self.pod.root)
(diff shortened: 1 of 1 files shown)


grow/deployments/destinations/webreview_destination.py
```diff
@@ -100,7 +100,7 @@ def deploy(self, *args, **kwargs):
                 url = finalize_response['fileset']['url']
                 # Append the homepage path to the staging link.
                 result = urlparse.urlparse(url)
-                if not result.path and self.pod.get_home_doc():
+                if not result.path and self.pod and self.pod.get_home_doc():
```
jeremydw: Would `self.pod` ever be None here? Thanks for the patch.

rspier: Yes.  It can happen if you don't have a webreview deployment in your podspec.

  File "snip/src/grow/bin/../grow/deployments/destinations/webreview_destination.py", line 103, in deploy
    if not result.path and self.pod.get_home_doc():
AttributeError: 'NoneType' object has no attribute 'get_home_doc'

jeremydw:
Thanks!

PR #367, opened 2017-02-07
Dependency Generation for `g.doc` and `g.docs`

Zoramite:
To be able to do incremental builds effectively we need to understand which other files have a dependency on specific files.

This is the beginning of that work allowing for the `g.doc` and `g.docs` template tags to report which document or collection is being used during a render operation.

### DIFF
--- a/grow/pods/dependency.py
+++ b/grow/pods/dependency.py
@@ -36,8 +36,10 @@ class DependencyGraph(object):
     def export(self):
         result = {}
 
-        for key, value in self._dependencies.iteritems():
-            result[key] = list(value)
+        for key in sorted(self._dependencies):
+            values = list(self._dependencies[key])
+            values.sort()
+            result[key] = values
 
         return result
 
--- a/grow/pods/dependency.py
+++ b/grow/pods/dependency.py
@@ -15,9 +15,18 @@ class DependencyGraph(object):
         self._dependents = {}
         self._dependencies = {}
 
+    def add_all(self, data):
+        """Add all from a data source."""
+
+        for key, value in data.iteritems():
+            self.add_references(key, value)
+
     def add_references(self, source, references):
         """Add references made in a source file to the graph."""
-        self._dependencies[source] = references
+        if not references:
+            return
+
+        self._dependencies[source] = set(references)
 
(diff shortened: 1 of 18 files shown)


Zoramite:
Thinking about this more, I am going to move the dependency graph to be attached to the pod (similar to pod routes) and make it lazy load.

This will keep the `render` clean and make it easier to interface with the yaml constructors later.

grow/pods/dependency.py
```diff
@@ -0,0 +1,74 @@
...
+
+class BadFieldsError(Error, ValueError):
+    pass
+
+
+class DependencyGraph(object):
+    def __init__(self):
+        self.reset()
+
+    def add_all(self, data):
```
jeremydw: What format is `data`? The convention we've used typically for naming is something like `file_ids_to_docs` (if `data` is a mapping of file IDs to documents, for example).

jeremydw: ... or, can it be anything? :)

grow/pods/pods.py
```diff
@@ -40,15 +41,14 @@ class PodDoesNotExistError(Error, IOError):
...
-class PodSpecParseError(Error):
-    pass
-
-
 # TODO(jeremydw): A handful of the properties of "pod" should be moved to the
 # "podspec" class.
 
 class Pod(object):
 
+    FILE_PODCACHE = 'podcache.yaml'
```
jeremydw: To avoid confusion with `podspec.yaml` I'd remove the `pod` from the name. It's been really nice to autocomplete `po-<TAB>` so far and I don't want to draw too much attention to this file since people won't be regularly interacting with it. Thoughts?

Zoramite: What if we change it to `.podcache.yaml`?

grow/pods/dependency.py
Zoramite: Done, thanks.

grow/pods/pods.py
jeremydw: That is sounding good to me.

Zoramite: Done, thanks.

Issue #95, opened 2015-08-03
Unable to access localized lists of strings.


meizon:
If I put a list of localized strings in YAML:

```
flavors@:
  - lemon
  - mango
  - raspberry
```

I am unable to access these using `doc.flavors` instead I have to do `doc['flavors@']`

jeremydw:
This was (finally) fixed in https://github.com/grow/grow/commit/ec4f063dca31c5b6cedcfe7e24f25b2241deba75

**Please note**! If you were previously using `{{doc['flavors@']}}` before to access strings from a list whose key ends in `@`, you'll now have to use `{{doc.flavors}}` as normal.

@stucox and @meizon I'd like your opinion on this as it's a breaking change. Unfortunately Jinja silently fails if you access `{{doc['flavors@']}}` and `flavors@` doesn't exist in the key.

So we have two options, I think, and neither are ideal but let me know what you think.

1. Be backwards compatible on this by duplicating any tagged lists that may exist in the yaml. This isn't ideal from a performance or code simplicity standpoint but it is doable. The content file would then have two entries per tagged list -- one with the tag and one without the tag.
1. Let this break and communicate in the release notes and email lists about this change.

jeremydw:
Okay, I was able to fix this with (1). Our fix is now backwards compatible with the previous behavior, and leaves tagged lists as tagged (but also duplicates them, so you can access the data in your example as `{{doc.flavors}}` instead of `{{doc['flavors@']}}`.

meizon:
I assume there will be a deprecation warning / policy for removing the doc['flavors@'] method?

jeremydw:
We're actually backwards compatible now (and I verified with existing sites), so deprecation warning isn't necessary at the moment since the previous behavior will continue to work. :)

PR #372, opened 2017-02-27
Lazy Loading Document

Zoramite:
Reworking a lot of the underlying parts of the document to make it as lazy loading as possible.

Builds strongly on the pod cache for caching the front-matter for each document and only loading from the file when the content of the document is needed. The caching is not currently being written during build, will do that as a separate PR.

The largest changes involve splitting the `formats.py` functionality up into discrete units:

  - `document_format.py`: Handle document formats by filetype (ex: html, markdown, yaml) and adjusts the processing.
  - `document_front_matter.py`: Deals specifically with handling the front matter of a document. Namely caching and parsing.
  - `document_fields.py`: Handles untagging the fields and will allow for some smarter property handling (ex: backwards compatability) in the future (ex: #95 can be updated to use this class and not have duplicate key/value pairs)

This change also removes the ability to do multi-locale documents as a single file. There is a `grow convert` commande  that will automate conversion from single files into multiple files.


grow/commands/convert.py
```diff
@@ -0,0 +1,20 @@
...
+
+@click.command()
+@click.argument('pod_path', default='.')
+@click.option('--version', help='What version to convert files to.')
+def convert(pod_path, version):
+    """Converts pod files from an earlier version of Grow."""
+    root = os.path.abspath(os.path.join(os.getcwd(), pod_path))
+    pod = pods.Pod(root, storage=storage.FileStorage)
+
+    if version == '0.1.0':
```
jeremydw: I wonder if we should be more declarative about this command and use names for the conversion methods rather than versions. For example, if a newer version of Grow detects a known-kind of parse error with YAML due to the removal of the split document feature, we could advise people to run `grow convert split-documents` (or some better name) rather than the version? That might be more clear for people?

Zoramite: That sounds like a good idea to me. I'll update it to use strings for the conversion.

grow/pods/collection_cache.py
```diff
@@ -0,0 +1,98 @@
...
+            doc.pod_path, doc._locale_kwarg)
+        self._cache[col.collection_path]['docs'][cache_key] = doc
+
+        # Recache with the real locale once we have stored it.
+        new_cache_key = CollectionCache.generate_cache_key(
+            doc.pod_path, doc.locale)
+        if cache_key != new_cache_key:
+            self._cache[col.collection_path]['docs'][new_cache_key] = doc
+
+    def delete_by_path(self, path):
```
jeremydw: This is referring just to the cache and not anything on the filesystem - right? If so, might be better to use "remove" rather than "delete". E.g. "add/remove".

grow/pods/document_fields.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+from boltons import iterutils
+import re
+
+
+LOCALIZED_KEY_REGEX = re.compile('(.*)@([^@]+)$')
+
+
+class DocumentFields(object):
```
jeremydw: This is great, and a long time coming. :) I wonder if we should just call it "Fields"? To keep things simple...

grow/pods/collection_cache.py
Zoramite: Sounds good, renamed to remove to not give confusing results.

grow/pods/document_fields.py
Zoramite: I was trying to keep all the document sub parts starting with document.

jeremydw: Doubtful - I just liked the idea of shorter file/object names. That said, encapsulating everything under `document_` sounds just fine if you think it'll be a good strategy to go with to keep things organized moving forward as we continue to work on this project.

jeremydw: One more thing – collection `_blueprint.yaml` files can technically have fields as well (e.g. titles, various metadata, collection-level content). So, just something to think about there.

grow/pods/pods.py
```diff
@@ -85,13 +88,17 @@ def _normalize_path(self, pod_path):
...
         podcache_file_name = '/{}'.format(self.FILE_PODCACHE)
         if not self.file_exists(podcache_file_name):
             return
         try:
-            return utils.parse_yaml(self.read_file(podcache_file_name))
+            with timer.Timer() as t:
+                # Do not use the utils.parse_yaml as that has extra constructors
+                # that should not be run when the cache file is being parsed.
+                parsed = yaml.load(self.read_file(podcache_file_name)) or {}
+            print "=> elasped _parse_cache_yaml: %s s" % t.secs
```
jeremydw: Should this make it into the branch?

grow/pods/routes.py
```diff
@@ -186,6 +186,7 @@ def match_error(self, path, status=404):
 
     def reset_cache(self, rebuild=True, inject=False):
         if rebuild:
+            self.pod.podcache.reset()
             self._build_routing_map(inject=False)
```
jeremydw: Just a note - I can't remember all the specifics around the routing map cache but that is something we could probably move to the new podcache as well eventually.

grow/pods/pods.py
Zoramite: Thanks, I'll take that out.

grow/pods/routes.py
Zoramite: Sounds like a good idea. I think with the improvements to the cache file it may not be an issue. I'm going to do some testing next to see what change to the route loading it makes to have the cache file. So far the changes don't actually use the cache file.

PR #375, opened 2017-03-10
Implement grow extract --audit.

jeremydw:
Provides the ability to audit content files for untagged strings.


jeremydw:
FYI @mhchen this PR fixes the issue you ran into this morning with Grow attempting to extract non-string values from yaml into PO files.

mhchen:
Thanks @jeremydw!

PR #376, opened 2017-03-11
Footnotes

Zoramite:
Adding built-in support for Footnotes.

The footnotes are specific to each page render so they will not have artifacts between pages.

Footnotes can be configured in the podspec to provide custom symbols, override to use only numbers, and provide a custom regex for locales that should be numeric instead of symbols.


grow/pods/footnotes.py
```diff
@@ -0,0 +1,89 @@
...
+    u'2': u'²',
+    u'3': u'³',
+    u'4': u'⁴',
+    u'5': u'⁵',
+    u'6': u'⁶',
+    u'7': u'⁷',
+    u'8': u'⁸',
+    u'9': u'⁹',
+}
+NUMERIC_LOCALES_REGEX = re.compile(r'_(DE|CA)$', re.IGNORECASE)
```
jeremydw: Nice! The "DE|CA" thing you're basing this on was a project-specific requirement.

Although I'm operating on good information that German sites should use numerical symbols by default, I would probably make this a little more configurable for users by moving this out into a constant such as `NUMERIC_TERRITORIES` so that people can override this on a per-project basis.

grow/pods/footnotes.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def numberic_symbol_generator():
+    index = 1
+    while True:
+        yield ''.join([NUMERICAL_SYMBOLS[char] for char in str(index)])
+        index += 1
+
+
+class Footnotes:
```
jeremydw: `class Footnotes(object)` for consistency

grow/pods/rendered.py
```diff
@@ -49,18 +50,33 @@ def render(self, params, inject=True):
...
+        notes = footnotes.Footnotes(
+            locale, symbols=symbols, use_numeric_symbols=use_numeric_symbols,
+            numeric_locales_pattern=numeric_locales_pattern)
+
         try:
             kwargs = {
                 'doc': self.doc,
                 'env': self.pod.env,
-                'podspec': self.pod.get_podspec(),
+                'footnotes': notes,
```
jeremydw: I think this should probably be a property on the `Document` object to avoid overloading our global with extra built-in vars. The footnotes are also document-specific so I think that would make sense.

Zoramite: They cannot be part of the document because they would be cached between renders.

grow/pods/footnotes.py
Zoramite: Fixed.

grow/pods/rendered.py
jeremydw: Ah, that's right. Just read through your messages on Slack as well. It's a good explanation. I think this makes sense. :) Keep as-is unless we think of something better!

grow/pods/footnotes.py
Zoramite: This is overriddable using the configuration.

ex:

    footnotes:
      numeric_locales_pattern: '_US$'

grow/pods/rendered.py
Zoramite: We could do something like 'utils' as a dict in the namespace? Then have `{{ utils.footnote.add('...') }}` ?

grow/pods/footnotes.py
Zoramite: Is the CA requirement not a standard thing? Or is it ok to use symbols in CA territories?

PR #386, opened 2017-04-05
Babel number filters

Zoramite:
Creating better babel filters. Making the date filters default to using the current doc's locale and adding number/currency/percent/decimal filters from babel.


jeremydw:
Looking good - this means that if the `locale` kwarg is not specified it's inferred from the current `doc`?

Zoramite:
Correct. It will use the current doc's locale if there is not a specified locale which makes it easier to use.

PR #392, opened 2017-04-09
Control Dev Server File Watching and Routes Rebuilding

Zoramite:
Several improvements to how the routes are handled on the dev server. The complete `routes` are seldom completely rebuilt. Instead the dependency graph is used to only update the routes as needed.

The document and collection cache are also correctly cleared when the file or a dependency are changed. This prevents weird states where the file is changed, but Grow uses a cached version.

This should vastly speed up the rendering time on the dev server, especially on large projects.


grow/pods/pods.py
```diff
@@ -290,6 +289,41 @@ def file_size(self, pod_path):
         path = self._normalize_path(pod_path)
         return self.storage.size(path)
 
+    def file_updated(self, pod_path):
```
jeremydw: Maybe this should be `on_file_updated` to better indicate it's a callback function versus our other functions like `file_exists`?

grow/preprocessors/file_watchers.py
```diff
@@ -1,12 +1,11 @@
-from . import routes_cache
 from . import translation
 from watchdog import events
 from watchdog import observers
 from xtermcolor import colorize
 
 
 class PodspecFileEventHandler(events.PatternMatchingEventHandler):
```
jeremydw: This should just be `PodFileEventHandler` now if it's not specific to podspec.

grow/pods/pods.py
Zoramite: Thanks. Updated to `on_file_changed`.

Zoramite:
@jeremydw these changes do speed it up. Before it would take about 1.5-2 seconds for a change (on a good sized project) to be shown after changing the file since the entire routes would need to be regenerated before the changes would be shown.

Most of this was actually from the `RenderController` keeping a static copy of the doc. With these changes I can't refresh it fast enough manually after a change to see a delay in the update. Also, the only time the routes are rebuilt is when one of the serving paths change on the dependents which works out really nicely.

Zoramite:
As a little test, it now takes about 9-15 ms on the medium sized site that I am using to test for the affected files to be flushed. So if you refresh the page within 15 ms of changing the file you might see the older file 😈

jeremydw:
Excellent. >:) Let's do it. Thanks Randy - this is just what we needed.

PR #408, opened 2017-05-05
Pass Grow environment name to preprocessors.

rspier:
Use the deployment's environment for earlier steps (such as preprocessing and stats) and not just deployment.

Exposes the name as the GROW_ENVIRONMENT_NAME environment name.


jeremydw:
Thanks for the contribution!

Issue #401, opened 2017-04-17
Warning when checking for updates


jeremydw:
When using `grow run`, the built version of Grow is unable to check for updates and shows the following warning:

```
[Errno socket error] [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:661) Unable to check for the latest version: [Errno socket error] [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:661)
```

coryphwang:
I've run into the exact same error message. Any ideas about what we can do to resolve this?

jeremydw:
Yep - we'll have this fixed for the next release. It only happens with the binary built on Circle CI, on Mac, when the development server checks for updates. It is fine to ignore the warning. Sorry about the log spam, though!

PR #415, opened 2017-05-25
Fix the SSL errors when checking for updates.

Zoramite:
This may work because requests is meant to use a cacerts.txt file bundled with Grow and urllib may not do that.


Zoramite:
Fixes #401

Issue #417, opened 2017-05-26
Update installer to check multiple bash files for grow alias


Zoramite:
Currently the installer checks only one bash config file for a grow alias. An improvement would be to search all existing config files for the alias to make sure they are not conflicting aliases.

Additionally it should check for multiple occurrences of the grow alias across all the config files to show a warning if it exists in multiple places.

jeremydw:
How does the new Google Cloud SDK work? The original code for Grow was based on how gcloud registered itself. There have been several changes/updates to gcloud SDK since, so I'm wondering if we can look at how gcloud works these days and do that too.

Zoramite:
Looks like gcloud does this:

```bash
# The next line updates PATH for the Google Cloud SDK.
if [ -f /Users/randy/google-cloud-sdk/path.bash.inc ]; then
  source '/Users/randy/google-cloud-sdk/path.bash.inc'
fi
```

Zoramite:
Also, their install script is very complex, I was digging into a bit. I don't think we'll be able to copy exactly what they do, but I think I have a nice solution for now.

jeremydw:
Sounds good, I just took a look at your solution and it looks like it'll cover the situations we care about when using the installer, so that's looking good.

Issue #416, opened 2017-05-26
Add installer check for incorrect grow alias binary


Zoramite:
When installing grow the installer checks the alias file for an existing alias before adding a new one.

In the case where it finds the alias it should be checked to make sure the path matches the binary being installed.

If the alias does not match a warning should be shown to let the user know that the alias is not pointing to the latest binary.

PR #418, opened 2017-05-26
Updating the install script to be more aware of existing aliases.

Zoramite:
If there are aliases that conflict an error message is also shown.

Also checks multiple alias files to check for existing alias.

Part of #417
Part of #416

### DIFF
--- a/install.py
+++ b/install.py
@@ -17,6 +17,10 @@ import zipfile
 
 DOWNLOAD_URL_FORMAT = 'https://github.com/grow/grow/releases/download/{version}/{name}'
 RELEASES_API = 'https://api.github.com/repos/grow/grow/releases'
+ALIAS_FILES_LINUX = ['.bash_aliases', '.bash_profile', '.profile', '.bashrc']
+ALIAS_FILES_MAC = ['.bash_profile']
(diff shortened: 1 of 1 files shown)


install.py
```diff
@@ -151,9 +181,9 @@ def install(rc_path=None, bin_path=None, force=False):
     fp.write('\n' + alias_comment + '\n')
     fp.write(alias_cmd)
     fp.close()
-    hai('{blue}[✓]{/blue} {green}Created "grow" alias in:{/green} %s' % rc_path)
+    hai('{blue}[✓]{/blue} {green}Created "grow" alias in:{/green} {}', rc_path)
     hai('{green}All done. To use Grow SDK...{/green}')
-    hai(' ...reload your shell session OR use `source {}`,'.format(rc_path))
+    hai(' ...reload your shell session OR use `source {}`,', rc_path)
```
jeremydw: The new gcloud SDK has the user do this to reload their shell instead:
```
exec -l $SHELL
```

WDYT of that approach, instead of telling people to resource a file? (Those instructions came from here - https://cloud.google.com/sdk/downloads#mac)

Zoramite: I like doing `source`, it seems like others might find it less threatening than `exec`?

PR #420, opened 2017-05-30
Improvements to the Google Sheets translation formatting

Zoramite:
When updating the meta information for a sheet:

  * The source locale is no longer uploaded as a sheet.
  * Missing translations are conditionally formatted to a light magenta color.
  * Source language cells are locked with a warning.
  * Description cells are locked with a warning.

### DIFF
--- a/grow/translators/google_sheets.py
+++ b/grow/translators/google_sheets.py
@@ -338,7 +338,7 @@ class GoogleSheetsTranslator(base.Translator):
 
         return requests
 
-    def _generate_style_requests(self, sheet_id, sheet):
+    def _generate_style_requests(self, sheet_id, sheet=None):
         formats = {}
 
         formats['header_cell'] = {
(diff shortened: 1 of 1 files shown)


grow/translators/google_sheets.py
```diff
@@ -396,10 +425,81 @@ def _generate_style_requests(self, sheet_id):
...
+        # Protect the location values.
+        requests += self._generate_style_protected_requests(sheet_id, sheet, {
+            'protectedRangeId': sheet_id + 1000002,  # Keep it predictble.
+            'range': {
+                'sheetId': sheet_id,
+                'startColumnIndex': 2,
+                'endColumnIndex': 3,
+                'startRowIndex': self.HEADER_ROW_COUNT,
+            },
+            'description': 'Description strings can only be edited in the source files.',
```
jeremydw: Should be something more like "Source strings can only be edited in source code." -> since the developer is the person who uploads this (or CI uploads this), and non-technical translators will be the ones seeing these settings, most likely.

grow/translators/google_sheets.py
```diff
@@ -351,9 +355,30 @@ def _generate_style_requests(self, sheet_id):
...
+                'green': 0.95,
+                'blue': 1
+            }
+        }
+
         formats['wrap'] = {'wrapStrategy': 'WRAP'}
 
         requests = []
+
+        # TODO Figure out how to be smarter about matching conditional formatting.
```
Zoramite: True, it just bugs me because if anyone wants to do something custom to their sheet we wipe it out. Not that it is always bad, just annoying to me that it can't be more precise :)

Issue #419, opened 2017-05-26
Add fields to sheets translator


stevenle:
At the moment, the sheets translator only uploads the string, translation, and file location. I have a use-case for a project that requires a "key" and translator comments. For the "key", would be good enough to use some hash of the string. 

It'd be cool to be able to configure the sheets translator to optionally add additional fields, maybe something like:

```
translators:
  - service: google_sheets
    acl:
    - domain: blinkk.com
      role: WRITER
    extra_fields:
    - md5_hash
    - translator_comments
```

LMK what you think.

Zoramite:
I like the idea of having the translator comments added as part of the spreadsheet.

I shy away from the idea of a hash in grow since that seems to be a very specific use case. For instance, you could do [something like this](https://stackoverflow.com/questions/7994410/hash-of-a-cell-text-in-google-spreadsheet) in an additional column for a hashed key that then would not need to be part of grow and would still be very useable since grow should not touch columns outside of the normal columns.

PR #421, opened 2017-05-30
Adding translator comments to the spreadsheet.

Zoramite:
Adds a protected column for any translator comments to be shown by the translation in the 3rd column.

Part of #419

### DIFF
--- a/grow/translators/google_sheets.py
+++ b/grow/translators/google_sheets.py
@@ -49,7 +49,7 @@ class GoogleSheetsTranslator(base.Translator):
     KIND = 'google_sheets'
     HEADER_ROW_COUNT = 1
     # Source locale, translation locale, message location.
-    HEADER_LABELS = [None, None, 'Location']
+    HEADER_LABELS = [None, None, 'Comments', 'Location']
     has_immutable_translation_resources = False
     has_multiple_langs_in_one_resource = True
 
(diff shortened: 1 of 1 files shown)


jeremydw:
Location will be in every sheet, but translator comments will only be on sheets where the developer has left comments for translators. Given this, should Comments be the last column?

stevenle:
Thanks for making the change. I'll leave it up to @jeremydw to decide whether this is the right user behavior, but for what it's worth, I'd also be happy to have the translator comments appended directly to the "Location" column in the same way the po files have it.

Also - just checking, did you verify backwards-compatibility for existing sheets? From a quick skim of the code, I can't tell if columns are updated using a key-value mapping or if it's done by column index, if the latter, updating existing sheets might insert the translator comments into the wrong column.

Zoramite:
@jeremydw My thoughts for having it be in third column instead of forth, is that if there are comments, then they are more important to the translators (who the sheet is made for) than the file that it is in.

@stevenle I think it would be better to be a separate column so that it doesn't confuse translators having the source information there as well. I think the translators would care more about the comment than the source file. And yes, it works in a backwards compatible way. :)

jeremydw:
The comments are indeed important, but something that happens really frequently is users need to find + audit all strings used on a particular page, so it turns out the location information is really important. For example, a translator asks me "how do I find the translations corresponding to the <A> page?" and I say "search for "<A>" and then filter the view by cells that contain "<A>.yaml".

This flow is important for users - especially when they need to update just a handful of strings within a file that has dozens of pages and hundreds of strings.

Zoramite:
Yeah, I think being able to see the source is still important, I thought that seeing the comments would be more helpful for normal translations. Having it be one column over wouldn't prevent the same filtering that they do now.

How about we add in a filter view that would allow them to filter down by the page? I would need to look into how to use it easily, but it might be a better way for finding translations by source?

Zoramite:
I've created #423 to explore using a filter view to be able to do the filtering by page in a way that doesn't affect everyone using the doc.

PR #427, opened 2017-06-02
Per content location Filter Views for Google Sheets Translation

Zoramite:
When running the `grow upload_translations --update-meta` the filter views will be updated for each content path.

image

### DIFF
--- a/grow/translators/google_sheets.py
+++ b/grow/translators/google_sheets.py
@@ -55,6 +55,7 @@ class GoogleSheetsTranslator(base.Translator):
     HEADER_ROW_COUNT = 1
     # Source locale, translation locale, message location.
     HEADER_LABELS = [None, None, 'Extracted comments', 'Reference']
+    SHEETS_BASE_URL = 'https://docs.google.com/spreadsheets/d/'
     has_immutable_translation_resources = False
     has_multiple_langs_in_one_resource = True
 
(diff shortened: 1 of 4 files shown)


grow/ui/templates/ui.html
```diff
@@ -21,7 +21,7 @@
 <script>
   grow.ui.main({
     docPodPath: '{{doc.pod_path}}',
-    injectEditUrl: '{{inject_edit_url or ''}}',
-    injectTranslateUrl: '{{translator_edit_url or ''}}'
+    {% if injectEditUrl %}injectEditUrl: '{{inject_edit_url|safe}}',{% endif %}
```
jeremydw: `if inject_edit_url`

grow/translators/google_sheets.py
```diff
@@ -382,11 +388,83 @@ def _generate_create_sheets_requests(self, catalogs, source_lang):
...
+        # Find all the unique resource locations.
+        for message in catalog:
+            if not message.id:
+                continue
+
+            for raw_location in message.locations:
+                location = raw_location[0]
+                if not location in location_to_filter_id:
+                    # Try to match up with the document hash value.
+                    location_to_filter_id[location] = self._trim_hash(
```
jeremydw: Can you move this to a function, something like `generate_filter_id` with args location and lang? Then we can reuse it easily for creating the per-document edit URL.

jeremydw: Maybe `create_filter_id` would be better. ;)

PR #445, opened 2017-06-16
Fix issue with grow install

jeremydw:
Fix chicken and egg situation with requiring extensions to being installed before being able to use grow install to install extensions. Solidify the directory 'extensions' as the canonical place for Grow-managed extensions to be installed. This means that 'extensions' should be .gitignored.

This will add support for extensions to be installed from our `grow-ext-*` repos by using `grow install`, fixing https://github.com/grow/grow/issues/293


grow/pods/pods.py
```diff
@@ -25,6 +25,21 @@
 from . import static
 from . import storage
 from . import tags
+from ..preprocessors import preprocessors
```
Zoramite: Want to check these? I think they are from before the pylint.

jeremydw: Fixed - thanks!

PR #449, opened 2017-06-20
Manifest generation

chargrizzle:
Command to generate a manifest using list_statics, this is useful if you have a build pipeline that needs access to the static paths.

### DIFF
--- a/grow/commands/__init__.py
+++ b/grow/commands/__init__.py
@@ -4,6 +4,7 @@ from . import deploy
 from . import download_translations
 from . import extract
 from . import filter
+from . import generate_manifest
 from . import import_translations
 from . import init
 from . import install
@@ -23,6 +24,7 @@ def add(group):
     group.add_command(download_translations.download_translations)
     group.add_command(extract.extract)
     group.add_command(filter.filter)
+    group.add_command(generate_manifest.generate_manifest)
     group.add_command(import_translations.import_translations)
     group.add_command(init.init)
     group.add_command(machine_translate.machine_translate)
--- a/grow/commands/generate_manifest.py
+++ b/grow/commands/generate_manifest.py
@@ -1,0 +1,20 @@
+from grow.pods import pods
+from grow.pods import storage
+import click
+import os
+import json
+
+
+@click.command()
+@click.argument('statics_path')
+@click.argument('pod_path', default='.')
+def generate_manifest(statics_path, pod_path):
+    """Generates a g.static manifiest given a pod path."""
+    root = os.path.abspath(os.path.join(os.getcwd(), pod_path))
+    pod = pods.Pod(root, storage=storage.FileStorage)
+    try:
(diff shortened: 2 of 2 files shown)


jeremydw:
Thanks for the PR! Wondering... should this be a flag to the existing command, `grow routes`? `grow routes` shows all routes that are built by the site; rendered routes and static routes included. 

I wonder if we should add an `output_style` flag and a `type` flag (maybe named something different?) to support the functionality in the existing command?

chargrizzle:
Sure! I'll add it to the existing command and we can get everything to dump as JSON

jeremydw:
Awesome. `output_style=json` seems like it would make a lot of sense to me.

chargrizzle:
Do you think it makes sense to just use stdout for this? The user could just pipe into a file if needed.

jeremydw:
Yes; the `click` command line utility we're using supports a way to do this. It should be able to detect if the user is piping something to stdout and do the right thing. If the user isn't piping anything then it should display in a pager.

jeremydw:
Notes:

http://click.pocoo.org/5/utils/#printing-to-stdout
http://click.pocoo.org/5/utils/#pager-support

Zoramite:
To be consistent with how much we use yaml files, maybe have the default output be yaml and allow the `output_format=json` flag to output it as json?

And I like the idea of just piping the output to a file instead of specifying an output file as an argument.

Zoramite:
@jeremydw  for separating out the static from the normal routes, what about doing a `filter=static` flag? Or would `type=static` be better?

chargrizzle:
Do we want to expand all the routes here? 

For example the current route output is something like:
```
<Static(format='/source/images/{filename}')>
  /static/images/1.png
  /static/images/2.png
  /static/images/3.png
```
Would expected behaviour output individual maps?
```
{
 "/source/images/1.png":" /static/images/1.png",
 "/source/images/2.png":" /static/images/2.png",
 "/source/images/3.png":" /static/images/3.png"
}
```
or combined?
```
{
"/source/images/{filename}": ["/static/images/1.png", "/static/images/2.png", "/static/images/3.png"]
}
```
in which case would the full class repr still be the key? 
```
"<Static(format='/source/images/{filename}')>" or "/source/images/{filename}"?
```

Also for routes would it make more sense to use the doc as the key or the view? Maybe something that needs to be added as an option?

Zoramite:
```json
{
"/source/images/{filename}": ["/static/images/1.png", "/static/images/2.png", "/static/images/3.png"]
}
```

Would be the better format for structuring since it adds the value of understanding which route is generating the filenames. And there is no need to use the entire `repr` for the key, just the format from the route should be enough.

Zoramite:
I'm going to remove this for now. We now have the `routescache.json` file that contains all of the routing information from a build. So we can use that as a form of a manifest that can be parsed and used in a pipeline.

PR #450, opened 2017-06-21
Pod Object Cache

Zoramite:
Object cache for the pod.

Allows for extensions or features to use a cache either globally on the pod or using a named cache to make it easier to avoid collision.

Caches can also be created with the option of not being cleared on a normal `podcache.clear()` or being able to be persisted in the `.podcache.yaml` file.


grow/pods/podcache.py
```diff
@@ -41,12 +55,59 @@ def document_cache(self):
         """Cache for specific document properties."""
         return self._document_cache
 
+    @property
+    def object_cache(self):
+        """Global object cache."""
+        return self.get_object_cache(self.KEY_GLOBAL)
+
+    def create_object_cache(self, key, write_to_file=False, can_reset=True, values=None):
```
jeremydw: Why does the object cache have a distinction for writing to a file, but the others don't? Should we just make the podcache always used for persistent (file) storage and not have the user have to make a decision about it?

Zoramite: I am treating the object cache as defaulting to a runtime only cache with the option to persist it to a file. The main idea was to keep the cache file from getting bloated with transient objects. Especially if the authors don't realize that their cache is being persisted. Also, having a large cache file can slow down the startup time.

By making the option to persist it to a file default to `False`, the object will be cached during a single usage of the pod (ex a `run` or `build`), but if the author of an extension or something internal wants it to be written to the file it is a conscious choice.

We can also highlight in the documentation for the object cache that if the author is persisting to the file they are also responsible for cleaning up unused cache items and that failing to do so can lead to bloat in the cache file.

jeremydw: Makes sense - though instead of using the podcache a user that wants to cache stuff in runtime only could just use werkzeug's cache or some other caching? Or is the idea to centralize caching here and provide a canonical interface?

Zoramite: I was going for more of a canonical way of caching. Then we can document and it would allow the caches to be easily shared between extensions if they understood the cache structure that the other extension is using.

jeremydw:
Thanks! :) The interface will work well for the extensions I am working on.

PR #457, opened 2017-06-24
Fix issue with PyInstaller-based distribution and extensions that depend on stdlib

jeremydw:
No-op for the default distribution.

### DIFF
--- a/bin/grow
+++ b/bin/grow
@@ -5,7 +5,6 @@ import os
 import sys
 
 
-
 def is_packaged_app():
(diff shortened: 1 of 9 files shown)


bin/grow
```diff
@@ -5,6 +5,7 @@ import os
 import sys
 
 
+
```
Zoramite: Nit

jeremydw: Fixed - thanks!

Issue #444, opened 2017-06-13
Add an environment/deployment flag to grow run


uxder:
I would like to be able to pass environment variables via the` grow run` command.

Given something like: 

` grow run --env=debug`

I want to be able access that in my templates to show content only if that flag is passed:

```
{% if env.name == 'debug' %}
   <div>Shown only for enviroment debug<div>
{% endif %}
```

This could also be tied to deployments in podspec.yaml.

` grow run --deployment=growsdk`

Zoramite:
That is a great idea. The flag should specify the deployment configuration to run instead of arbitrary environment variables.

PR #465, opened 2017-07-01
Add support for @env.<name> conditional yaml syntax.

jeremydw:
Adds the ability to conditionally set parameters for YAML data (including in podspec) to configure things based on the environment name. See explanation in https://github.com/grow/grow/issues/464

Adds ability to do stuff like the following:

```
foo: bar-anywhere-but-prod
foo@env.prod: bar-in-prod
```

/cc @uxder @stevenle @kezlya

### DIFF
--- a/grow/pods/pods.py
+++ b/grow/pods/pods.py
@@ -59,6 +59,7 @@ class Pod(object):
                 and self.root == other.root)
 
     def __init__(self, root, storage=storage.auto, env=None, load_extensions=True):
+        self._yaml = utils.SENTINEL
         self.storage = storage
         self.root = (root if self.storage.is_cloud_storage
                      else os.path.abspath(root))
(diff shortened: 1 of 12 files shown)


jeremydw:
This PR also fixes #444 and adds support for `grow run --deployment=<deployment>` to make it easy to test the behavior of `grow build` using the dev server.

grow/pods/pods.py
```diff
@@ -125,6 +126,17 @@ def _parse_yaml(self):
                 raise PodDoesNotExistError('Pod not found in: {}'.format(path))
             raise podspec.PodSpecParseError('Error parsing: {}'.format(path))
 
+    def set_env(self, env):
+        if env.name:
+            untag = document_fields.DocumentFields.untag
+            content = untag(self.yaml, env_name=env.name)
+            self._yaml = content
```
Zoramite: I think this will cause issues if you ever change the environment more than once. The content will become filtered the first time that you `set_env()` and later attempts will always be the same since there are no longer any tagged versions.

Use the `_parse_yaml` (since it is memoized) as the source for all environment untagging instead of `self.yaml`?

jeremydw: Ah, great catch. Updated. Thanks!

PR #467, opened 2017-07-05
Use tagged fields to determine default_locale

### LINKED ISSUE #448: Locale value overrides with default_locale
I think there might be an issue with localized values and default_locale. In the following snippet, I'd expect "French value" to be output, but I think it actually outputs "Default value".

```yaml
# doc.yaml

$localization:
  default_locale: fr

foobar@: Default value
foobar@fr: French value
```

```jinja
{# view.html #}

{{_(doc.foobar)}}
```

One way I was able to quickly reproduce was to `grow init codelab ./my-codelab/` and then change `default_locale` and `tagline` in index.md.

jeremydw:
Fixes #448

### DIFF
--- a/grow/pods/routes.py
+++ b/grow/pods/routes.py
@@ -1,7 +1,8 @@
 """Routes for document path routing."""
 
 import collections
-import webob
+# NOTE: exc imported directly, webob.exc doesn't work when frozen.
+from webob import exc as webob_exc
 from werkzeug import routing
 from grow.common import timer
 from grow.common import utils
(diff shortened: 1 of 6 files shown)


grow/pods/routes.py
```diff
@@ -1,7 +1,8 @@
 """Routes for document path routing."""
 
 import collections
-import webob
+# NOTE: exc imported directly, webob.exc doesn't work when frozen.
```
Zoramite: Are these changes to the webob a new change or just unmerged changes?

jeremydw: A new change - I cleaned it up inside this PR too.

PR #478, opened 2017-07-20
Adding deep references for yaml in the `g.yaml` constructor. 

Zoramite:
Allows for loading a yaml document and then only using a specific key from the imported file.

Uses the following format the the yaml constructor:

`!g.yaml /content/shared/stuff.yaml?drill.down.to.stuff`

Which translates to `stuff['drill']['down']['to']['stuff']` being used as the value instead of the entire yaml file.

When there is no matching key in the file it returns a `None` value.


Zoramite:
@jeremydw Going to merge this feature for now to make a release. If we need to adjust it we can make a follow up release.

PR #479, opened 2017-07-21
Logging Updates

Zoramite:
Changing around the logging to try and prevent massive logging repeats when threaded.

### DIFF
--- a/grow/common/logger.py
+++ b/grow/common/logger.py
@@ -1,0 +1,15 @@
+"""Logging utilities for grow."""
+
+import logging
+
+def get_logger(name='pod'):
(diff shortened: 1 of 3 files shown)


jeremydw:
1. Not sure why this is failing GAE test - really bizarre. Are all our dependencies pinned?
1. Hm - the logger change I hope works. :\ On paper I feel like this is the same as it was originally (just a global logger in `pods.py`) - can you double check this is fixing the duplicate logging issues when using Kintaro preprocessor (and running the `preprocess` command for example?) I'll check myself tomorrow when I'm back from OOO.

Otherwise LGTM :)

PR #483, opened 2017-07-27
Incremental builds for `grow build`

### LINKED ISSUE #365: Support for incremental `content/` build.
**Goal:** Support incremental builds based on source file change.

Currently the build command builds the entire site. When running the dev server a specific URL can be recompiled at request. But there is not a way to do an incremental build.

The ideal flow for this option is to be able to update a source file. Run the build command with a specific flag and grow will recompile any of paths that rely upon the source file.

For instance, if just editing content for a page, then only one output file is affected. But if you were editing a header.yaml file than all the pages may need to be rebuilt.

Potential command line usage:

    grow build --file=/content/pages/home.yaml

The immediate implementation will only work for `/content` files.

*This is not meant to replace the local development server. Instead it is an ability that would be useful in a staging environment for editing content for non-developers.*

Zoramite:
Adds the ability to use the `--files=/content/...` flag to the build command and have it only build the content that is a dependent on the file.

Also updates the diff generated to not do deletes since there isn't enough information to know if files were deleted without a full build.

Fixes #365

### DIFF
--- a/grow/commands/build.py
+++ b/grow/commands/build.py
@@ -1,10 +1,12 @@
+"""Command for building pods into static deployments."""
+
+import os
+import click
 from grow.common import utils
 from grow.deployments import stats
 from grow.deployments.destinations import local as local_destination
 from grow.pods import pods
 from grow.pods import storage
-import click
-import os
 
 
 @click.command()
@@ -16,8 +18,8 @@ import os
 @click.option('--clear_cache',
               default=False, is_flag=True,
               help='Clear the pod cache before building.')
-@click.option('--files', help='Build only pages affected by content files.', multiple=True)
-def build(pod_path, out_dir, preprocess, clear_cache, files):
+@click.option('--file', '--pod_path', 'pod_paths', help='Build only pages affected by content files.', multiple=True)
+def build(pod_path, out_dir, preprocess, clear_cache, pod_paths):
     """Generates static files and dumps them to a local destination."""
     root = os.path.abspath(os.path.join(os.getcwd(), pod_path))
     out_dir = out_dir or os.path.join(root, 'build')
@@ -29,11 +31,11 @@ def build(pod_path, out_dir, preprocess, clear_cache, files):
     try:
         config = local_destination.Config(out_dir=out_dir)
         destination = local_destination.LocalDestination(config)
-        paths_to_contents = destination.dump(pod, files=files)
+        paths_to_contents = destination.dump(pod, pod_paths=pod_paths)
         repo = utils.get_git_repo(pod.root)
         stats_obj = stats.Stats(pod, paths_to_contents=paths_to_contents)
         destination.deploy(paths_to_contents, stats=stats_obj, repo=repo, confirm=False,
-                           test=False, is_partial=bool(files))
+                           test=False, is_partial=bool(pod_paths))
         pod.podcache.write()
     except pods.Error as e:
         raise click.ClickException(str(e))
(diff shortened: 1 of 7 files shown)


grow/commands/build.py
```diff
@@ -16,7 +16,8 @@
 @click.option('--clear_cache',
               default=False, is_flag=True,
               help='Clear the pod cache before building.')
-def build(pod_path, out_dir, preprocess, clear_cache):
+@click.option('--files', help='Build only pages affected by content files.', multiple=True)
```
jeremydw: I think this should be `file` to be consistent with other flags - then you can do `--file=<> --file=<> --file=<>` etc.

Zoramite: It caused some problems with the built-in file: `redefined-builtin Redefining built-in 'file'`

Zoramite: An alternative that wouldn't conflict may work better? `pod_path`?

jeremydw: I like `pod_path` because that's super clear (doesn't conflict with "path" which might mean URL path). It's just a bit verbose. I'm trying to see if click has a way to use an option with one name but the kwarg to the function be named something else...

Zoramite:
Let me know if there is anything else we need to update.

jeremydw:
Did some testing - and maybe found a bug. See traceback below -

```
$ ~/git/grow/scripts/grow build -np --file=/content/pages/phone.yaml 
  File "/Users/jeremydw/git/grow/scripts/../bin/grow", line 92, in <module>
    grow()
  File "/Users/jeremydw/git/grow/env/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jeremydw/git/grow/env/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/jeremydw/git/grow/env/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/jeremydw/git/grow/env/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/jeremydw/git/grow/env/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/jeremydw/git/grow/bin/../grow/commands/build.py", line 34, in build
    paths_to_contents = destination.dump(pod, pod_paths=pod_paths)
  File "/Users/jeremydw/git/grow/bin/../grow/deployments/destinations/base.py", line 231, in dump
    return pod.dump(pod_paths=pod_paths)
  File "/Users/jeremydw/git/grow/bin/../grow/pods/pods.py", line 217, in dump
    suffix=suffix, append_slashes=append_slashes, pod_paths=pod_paths)
  File "/Users/jeremydw/git/grow/bin/../grow/pods/pods.py", line 232, in export
    routes = grow_routes.Routes.from_docs(self, _gen_docs(pod_paths))
  File "/Users/jeremydw/git/grow/bin/../grow/pods/routes.py", line 46, in from_docs
    routes._build_routing_map_from_docs(docs)
  File "/Users/jeremydw/git/grow/bin/../grow/pods/routes.py", line 83, in _build_routing_map_from_docs
    self._recreate_routing_map()
  File "/Users/jeremydw/git/grow/bin/../grow/pods/routes.py", line 123, in _recreate_routing_map
    rules = [rule.empty() for rule in self._routing_rules]
AttributeError: 'NoneType' object has no attribute 'empty'
```

PR #487, opened 2017-07-31
Basic Untranslated Stats on Build

Zoramite:
Adds the ability to display stats on untranslated when doing a build using the `--display-untranslated` flag.

This is precursor to being able to prevent a deployment from being deployed when there are untranslated strings.

Part of #383


grow/commands/build.py
```diff
@@ -18,8 +18,11 @@
 @click.option('--clear_cache',
               default=False, is_flag=True,
               help='Clear the pod cache before building.')
-@click.option('--file', '--pod_path', 'pod_paths', help='Build only pages affected by content files.', multiple=True)
-def build(pod_path, out_dir, preprocess, clear_cache, pod_paths):
+@click.option('--file', '--pod-path', 'pod_paths', help='Build only pages affected by content files.', multiple=True)
+@click.option('--display-untranslated',
```
jeremydw: This is a great change. Thank you for working on this. Just a small nit on naming (know we talked about this but now that we thought of collecting untranslated messages into a PO file)...

Wonder if this flag should be simplified to `--show-untranslated` or `--collect-untranslated` (assuming it will eventually write the `untranslated.po` file?)

Zoramite: I've renamed it to `--locate-untranslated`. Then it is a bit more broad than just display and we can use it to create the `.po` file as needed.

PR #508, opened 2017-08-16
Removing Global Abuse in Dependency Tracking

Zoramite:
Currently the dependency tracking in jinja templates abuses the globals object which can cause issues in the future if we get ways to parallelize the rendering since it is a shared scope.

This updated method uses copied code from the jinja library with an added argument for passing the `context` to the `_get_template` method. If there is a context method for `_track_dependency` then it is called with the template value.

This allows for a normal render to pass in a method that can be called to set the dependency using a context variable instead of a jinja global variable overwrite and should be safe when doing multiple renders at one time.


grow/templates/jinja_dependency.py
```diff
@@ -0,0 +1,112 @@
+"""Track the jinja2 rendering dependencies."""
```
jeremydw: What does a developer need to know about the maintainability of `DepEnvironment`? Is there any way to simply have a loader to track the deps instead of having a custom Environment? It might be good to have a quick comment in the file about how to update this or why it was necessary. :)

Zoramite: Good point. I've added in some better explanation of why I am using that method and how to go about updating when updating jinja.

PR #509, opened 2017-08-22
Streaming Pod Builds

Zoramite:
Adding  'streaming' build for the pod. Instead of rendering all the paths into a full object, this changes the `export` and `dump` methods to use a generator to render the documents.

The diff and index building are also updated to work with the streaming content. The content is only stored when it is different from the existing content. This should use a lot less overall memory on large sites.

### DIFF
--- a/grow/pods/pods.py
+++ b/grow/pods/pods.py
@@ -248,8 +248,9 @@ class Pod(object):
         for output_path, rendered in self.export_gen(
                 suffix=suffix, append_slashes=append_slashes, pod_paths=pod_paths):
             yield output_path, rendered
-        # if self.ui and not self.is_enabled(self.FEATURE_UI):
-        #     output.update(self.export_ui())
+        if self.ui and not self.is_enabled(self.FEATURE_UI):
+            for output_path, rendered in self.export_ui_gen():
+                yield output_path, rendered
 
     def enable(self, feature):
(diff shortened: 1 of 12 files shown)


grow/pods/pods.py
```diff
@@ -234,21 +234,8 @@ def delete_files(self, pod_paths, recursive=False, pattern=None):
...
-        if self.ui and not self.is_enabled(self.FEATURE_UI):
-            output.update(self.export_ui())
-        return output
-
-    def enable(self, feature):
-        self._disabled.discard(feature)
-
-    def export(self, suffix=None, append_slashes=False, pod_paths=None):
-        """Builds the pod, returning a mapping of paths to content based on pod routes."""
+    def determine_paths(self, pod_paths=None):
```
jeremydw: `determine_paths_to_build` for clarity

grow/pods/pods.py
```diff
@@ -708,6 +679,36 @@ def read_yaml(self, path):
         untag = document_fields.DocumentFields.untag
         return untag(fields, env_name=self.env.name)
 
+    def render_paths(self, paths, routes, suffix=None, append_slashes=False):
+        """Builds the pod, returning a mapping of paths to content."""
```
jeremydw: Update docstring

PR #513, opened 2017-08-24
Add support for --path and -o to grow extract command.

jeremydw:


grow/pods/catalog_holder.py
```diff
@@ -2,6 +2,7 @@
 
 import cStringIO
 import collections
+import fnmatch
```
jeremydw: Done

PR #518, opened 2017-08-25
Google Sheets to Collection

### LINKED ISSUE #511: Multi-tab downloads for Google Sheets preprocessor
Currently, a user has to manually add an extra preprocessor configuration in `podspec.yaml` in order to download multiple tabs; in other words, one preprocessor per tab (`gid`).

- We should enhance the functionality of the preprocessor to add support for downloading multiple tabs at once.
- The user should be able to specify a path to download the tabs to.
- The default format should be `yaml` (it should convert the output to yaml) and it should write one file per tab. The name of the file should be the "slugified" version of the tab's title.
- If "all" gids are specified, it should skip downloading any tab that begins with a "_".

Sample configuration (I'm open to different names, these are just a suggestion):

```
preprocessors:
- kind: google_sheets
  id: <sheet id>
  gids: all  # Accepts "all" or a list of gids to download.
  path: /content/strings/  # Downloads one yaml per tab to this folder.
```

Zoramite:
Updating the google sheets preprocessor to be able to import sheets into a collection.

Fixes #511

### DIFF
--- a/grow/preprocessors/google_drive.py
+++ b/grow/preprocessors/google_drive.py
@@ -374,6 +374,10 @@ class GoogleSheetsPreprocessor(BaseGooglePreprocessor):
             logger=self.pod.logger)
 
         if self.config.path:
+            if format_as in ['list']:
+                self.pod.logger.info(
+                    'Cannot inject list formatted spreadsheet -> {}'.format(self.config.path))
+                return
             # Single sheet import.
             path, key_to_update = self._parse_path(self.config.path)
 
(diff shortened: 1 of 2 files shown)


grow/preprocessors/google_drive.py
```diff
@@ -161,74 +164,140 @@ def _update_node(root, part):
...
+
+            for gid in gids:
+                if gid_to_sheet[gid]['title'].strip().startswith('_'):
+                    continue
+                file_name = '{}.yaml'.format(
+                    utils.slugify(gid_to_sheet[gid]['title']))
+                output_path = os.path.join(collection_path, file_name)
+                self.pod.write_yaml(output_path, gid_to_data[gid])
+                self.logger.info(
+                    'Downloaded {} -> {}'.format(gid_to_sheet[gid]['title'], output_path))
```
jeremydw: Would be good to include the GID after the title, just in case the user wants to go and update the `gids` setting later.

E.g.

```Downloaded Title (GID) -> filename``

PR #520, opened 2017-08-25
Use the `_` as the ignore character for keys.

### LINKED ISSUE #519: format_as: strings for Google Sheets preprocessor
Add an additional `format_as` option to the preprocessor, should do the same thing as `format_as: map` but automatically suffix the keys with `@`

If a key is blank or starts with `_` it should not be downloaded.

Zoramite:
Fixes #519

### DIFF
--- a/grow/preprocessors/google_drive.py
+++ b/grow/preprocessors/google_drive.py
@@ -29,6 +29,7 @@ from . import base
 
 OAUTH_SCOPE = 'https://www.googleapis.com/auth/drive'
 STORAGE_KEY = 'Grow SDK'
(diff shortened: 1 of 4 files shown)


grow/preprocessors/google_drive.py
```diff
@@ -148,7 +148,7 @@ def _update_node(root, part):
         for row in reader:
             key = row[0]
             value = row[1]
-            if key.startswith('#'):
+            if key.startswith('_'):
```
jeremydw: Maybe do `key.startswith(('_', '#'))` to ignore both?

PR #524, opened 2017-08-30
Podcache write on dev

Zoramite:
Modifying Dev server to write the cache file when it detects changes to the dependencies while rendering.

### DIFF
--- a/grow/templates/filters.py
+++ b/grow/templates/filters.py
@@ -42,6 +42,11 @@ def deeptrans(ctx, obj):
     """Deep translate an object."""
     return _deep_gettext(ctx, obj)
 
+@jinja2.contextfilter
+def expand_partial(_ctx, partial_name):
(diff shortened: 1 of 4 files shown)


grow/templates/filters.py
```diff
@@ -42,6 +42,11 @@ def deeptrans(ctx, obj):
     """Deep translate an object."""
     return _deep_gettext(ctx, obj)
 
+@jinja2.contextfilter
+def expand_partial(_ctx, partial_name):
+    """Filter for expanding partial path from name of partial."""
+    return '/partials/{0}/{0}.html'.format(partial_name)
```
jeremydw: hmm... can you always assume there are two things to substitute in `partial_name`?

Everything else looks good. :)

Zoramite: It is using positional substitution so it comes out to be `/partials/name/name.html`

Issue #527, opened 2017-08-31
Deep Profiling


Zoramite:
It is difficult to understand which parts of grow are having performance impacts. We can get anecdotal scenarios from working with grow that in certain situations it is not running well. But it is difficult to pin down which things are causing the slowness.

One solution is to be able to do a form of deep profiling around core operations in grow so that we can get detailed timing information.

PR #533, opened 2017-09-05
Deep Profiling Timers

Zoramite:
Initial timers for measuring performance.

Part of #527


grow/commands/build.py
```diff
@@ -23,7 +25,10 @@
 @click.option('--locate-untranslated',
               default=False, is_flag=True,
               help='Shows untranslated message information.')
-def build(pod_path, out_dir, preprocess, clear_cache, pod_paths, locate_untranslated):
+@click.option('--profile',
```
jeremydw: Might be good to move this to a global flag so we can get profiles not on just `grow build` but other operations. For example, `grow stats` takes a while for big sites.

grow/commands/build.py
```diff
@@ -35,18 +36,20 @@ def build(pod_path, out_dir, preprocess, clear_cache, pod_paths, locate_untransl
...
-        destination = local_destination.LocalDestination(config)
-        destination.pod = pod
-        paths, _ = pod.determine_paths_to_build(pod_paths=pod_paths)
-        repo = utils.get_git_repo(pod.root)
-        stats_obj = stats.Stats(pod, paths=paths)
-        content_generator = destination.dump(pod, pod_paths=pod_paths)
-        destination.deploy(content_generator, stats=stats_obj, repo=repo, confirm=False,
-                           test=False, is_partial=bool(pod_paths))
-        pod.podcache.write()
+        with pod.profile.timer('grow_build'):
```
jeremydw: Hm I wonder if this can be built in to the top-level command?

Zoramite: I thought about this, but it would require that we create the pod and place it in the context and not create it in each command. I'll test it out and see how well it works.

Zoramite: I've dug into this idea and it doesn't quite work out. To make it work I would need to be able to access the arguments for the sub-command, which click is very explicit in not allowing. I've played around with a few ideas, but it has not worked.

The main issue right now is that we need to get access to the `pod_path` argument that is passed to the subcommand while in the main command. But that does not work for click and a couple of attempts at custom decorators has also not worked.

Zoramite:
I've moved the main grow command to the `grow/commands/` directory if you want to make sure it looks correct.

PR #534, opened 2017-09-05
Logging errors, but not stopping the entire file watching.

### LINKED ISSUE #528: Errors encountered by file watchers require grow restart
When a file watcher encounters an error, it seems to require a restart of the dev server for future YAML changes to be picked up.

```
  File "threading.py", line 801, in __bootstrap_inner
  File "watchdog/observers/api.py", line 199, in run
  File "watchdog/observers/api.py", line 368, in dispatch_events
  File "watchdog/events.py", line 454, in dispatch
  File "grow/preprocessors/file_watchers.py", line 23, in on_created
  File "grow/preprocessors/file_watchers.py", line 19, in handle
  File "grow/pods/pods.py", line 599, in on_file_changed
  File "grow/pods/pods.py", line 388, in get_doc
  File "grow/pods/collection.py", line 255, in get_doc
  File "grow/pods/documents.py", line 74, in __init__
  File "grow/common/utils.py", line 134, in validate_name
BadNameError: Names must be lowercase and only contain letters, numbers, backslashes, and dashes. Found: "/content/modules/foo copy.yaml"
```

Zoramite:
Fixes #528

### DIFF
--- a/grow/preprocessors/file_watchers.py
+++ b/grow/preprocessors/file_watchers.py
@@ -16,7 +16,14 @@ class PodFileEventHandler(events.PatternMatchingEventHandler):
 
     def handle(self, event=None):
         pod_path = event.src_path[len(self.pod.root):]
-        self.pod.on_file_changed(pod_path)
+        try:
+            self.pod.on_file_changed(pod_path)
+        except Exception as err:
+            # Avoid an inconsistent state where preprocessor doesn't run again
+            # if it encounters an exception. https://github.com/grow/grow/issues/528
+            text = colorize('Preprocessor error.', ansi=197)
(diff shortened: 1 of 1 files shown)


jeremydw:
Awesome, easy fix. I verified this works. Let's clean up the output though.

With the two `logging.exception`s we get the traceback twice and on top of that it's hard to see what broke. I might suggest an update to the following:

```
colored_pod_path = colorize(pod_path, ansi=197)
self.pod.logger.exception('Found an error -> {}'.format(colored_pod_path))
```

And that should make the offending file really easy to see inside the traceback.

I think `logging.exception` automatically shows the traceback AND the error message you specify, so no need to log out the traceback explicitly.

PR #549, opened 2017-09-18
Updating the untagging to support a params object.

### LINKED ISSUE #547: Refactor untagging of `env` to be more robust
Currently when untagging there is a special case of being able to use `@env.prod` style tags to denote the environment to match for untagging.

We should expand upon this to instead pass an object that can be used for the untagging.

For example:

```
# Current
untag(data, locale='en_US', env_name='prod')

# Suggested
untag(data, locale='en_US', {'env': 'prod'})
```

Then if we want to expand this later for other use cases we can follow the format for tags: `@key.value` (ex `{'env': 'staging'}` => `@env.staging`

Zoramite:
This allows for untagging with an arbitrary set of prefixed values.

For example, `params={'env': 'prod'}` would match `key@env.prod`.

Fixes #547

### DIFF
--- a/grow/pods/document_fields.py
+++ b/grow/pods/document_fields.py
@@ -47,7 +47,6 @@ class DocumentFields(object):
             if params:
                 param_regex = re.compile(
                     r'(.*)@({})\.([^@]+)$'.format('|'.join(params.keys())))
-                print param_regex.pattern
                 param_match = param_regex.match(key)
                 if param_match:
                     untagged_key, param_key, param_value = param_match.groups()
--- a/grow/pods/document_fields.py
+++ b/grow/pods/document_fields.py
@@ -4,8 +4,7 @@ from boltons import iterutils
 import re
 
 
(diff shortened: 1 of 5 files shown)


grow/pods/document_fields.py
```diff
@@ -41,15 +42,21 @@ def visit(path, key, value):
...
-                if not env_name or not re.match(env_regex, env_name):
-                    return False
-                updated_localized_paths.add((path, untagged_key.rstrip('@')))
-                return untagged_key, value
+
+            # Support <key>@<param key>.<param value>: <value>.
+            if params:
+                param_regex = re.compile(
+                    r'(.*)@({})\.([^@]+)$'.format('|'.join(params.keys())))
+                print param_regex.pattern
```
jeremydw: Errant print ;)

jeremydw:
This is a good change, thanks for adding it. Looking forward to seeing what other conditional data we might have. :)

PR #559, opened 2017-09-21
Adding ability for spreadsheets to be formatted as a `grid`.

Zoramite:
This takes the first column to use as an id and the remaining columns are made into an object.

For example:

```
----------------------------------
| ID      | key1      |  key2    |
----------------------------------
| test    | foo       | bar      |
| test2   | baz       | bam      |
----------------------------------
```

Becomes:

```
test:
  key1: foo
  key2: bar
test2:
  key1: baz
  key2: bam
```


jeremydw:
Thanks for the quick turnaround.

PR #579, opened 2017-09-30
Initial replacement of the `(path, content)` with `RenderedDocument`.

Zoramite:
Instead of returning a path and a content, this changes it to return a `RenderedDocument` object. This object will write the contents to a temp file then store the reference. Then when the content is needed at upload it can read the content back in from the temp file.

This will hopefully reduce the about of memory needed as the objects about the content are quite small.

### DIFF
--- a/grow/pods/rendered_document.py
+++ b/grow/pods/rendered_document.py
@@ -36,8 +36,8 @@ class RenderedDocument(object):
 
         self.hash = hashlib.sha1(content).hexdigest()
 
-        if tmp_dir:
-            with open(self.filename, "w") as tmp_file:
+        if self.tmp_dir:
+            with open(self._get_tmp_filename(), "w") as tmp_file:
                 tmp_file.write(content)
         else:
             self._content = content
--- a/grow/pods/rendered_document.py
+++ b/grow/pods/rendered_document.py
@@ -1,0 +1,14 @@
+"""Tracks the rendered document information."""
+
+import hashlib
+
+class RenderedDocument(object):
+    """Keeps track of the information for the rendered document."""
+
+    def __init__(self, path, content):
+        if isinstance(content, unicode):
+            content = content.encode('utf-8')
+
+        self.path = path
+        self.content = content
+        self.hash = hashlib.sha1(content).hexdigest()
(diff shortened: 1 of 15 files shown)


grow/pods/rendered_document.py
```diff
@@ -0,0 +1,39 @@
...
+            return self.content
+
+        with open(self.filename, "r") as tmp_file:
+            file_contents = tmp_file.read()
+            if isinstance(file_contents, unicode):
+                file_contents = file_contents.encode('utf-8')
+            return file_contents
+
+    @property
+    def filename(self):
```
stevenle: in future, this implementation might change where it's not file-based, so might be better to make this a protected or private property if it's only used within this class.

grow/pods/rendered_document.py
```diff
@@ -0,0 +1,39 @@
...
+        self.tmp_dir = tmp_dir
+
+        if tmp_dir:
+            with open(self.filename, "w") as tmp_file:
+                tmp_file.write(content)
+        else:
+            self._content = content
+
+    @property
+    def content(self):
```
stevenle: I wonder if we should prefer more of a file-based API so that we can eventually optimize further with buffered reads and writes. We don't necessarily need to do the latter now, but initially we could potentially support that in the future by having a few methods like:

- open -> returns a file descriptor (if no tmp_dir is provided, returns a cStringIO instance)
- write -> opens the file descriptor and writes to it
- read -> opens the file descriptor and reads the entire contents

this might change the initialization a bit, so instead of passing contents directly to rendereddoc, you'd do:

```
doc = RenderedDoc(path, tmp_dir=foo)
doc.write(content)
```

Then i think to the caller, it'll feel more clear that when they call `doc.read()` that it's being read from the filesystem, whereas accessing attributes like `doc.content` always feels like it's stored in memory to me.

Zoramite: I've updated it to use a `write` and `read` functions. For now I'm going to forgo the `open` until we actually need it.

grow/pods/rendered_document.py
Zoramite: Good point. I've made it a protected function instead.

PR #622, opened 2017-11-01
Improving the local file deployment by using copy instead of read/write.

### LINKED ISSUE #618: Optimize local file deployments
I noticed that for local file deployments, once the site is done building, it takes a while to copy files from the tmp dir into its destination. I wonder if it's because we're essentially doing `dst_file.write(src_file.read())` as opposed to letting the filesystem handle optimizations using `shutil.copyfile(src, dst)` (or even `shutil.move(src, dst)`).

Zoramite:
Great improvement for local deployments.

With read/write deployment: `Deploying: 4457/4457 (in 0:01:00.1)`
With copy deployment: `Deploying: 4457/4457 (in 0:00:01.1)`

🚀 

Fixes #618

### DIFF
--- a/grow/deployments/destinations/local.py
+++ b/grow/deployments/destinations/local.py
@@ -1,8 +1,11 @@
-from . import base
-from protorpc import messages
+"""Local deployment destination."""
+
+import os
 from grow.pods import env
 from grow.pods.storage import storage as storage_lib
-import os
+from grow.deployments.destinations import base
+from protorpc import messages
+from shutil import copyfile
 
 
 class Config(messages.Message):
@@ -36,10 +39,11 @@ class LocalDestination(base.BaseDestination):
 
     def write_file(self, rendered_doc):
         path = rendered_doc.path
-        content = rendered_doc.read()
         out_path = os.path.join(self.out_dir, path.lstrip('/'))
-        fp = self.storage.write(out_path, content)
-        fp.close()
+        if rendered_doc.filename:
+            copyfile(rendered_doc.filename, path)
+        else:
+            self.storage.write(out_path, rendered_doc.read())
 
     def prelaunch(self, dry_run=False):
         for command in self.config.before_deploy:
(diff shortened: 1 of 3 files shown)


grow/deployments/destinations/local.py
```diff
@@ -1,8 +1,11 @@
...
-from protorpc import messages
+"""Local deployment destination."""
+
+import os
 from grow.pods import env
 from grow.pods.storage import storage as storage_lib
-import os
+from grow.deployments.destinations import base
+from protorpc import messages
+from shutil import copyfile
```
stevenle: import module instead, e.g. `import shutil`

grow/pods/storage/file_storage.py
```diff
@@ -62,10 +62,9 @@ def write(cls, path, content):
                 pass
             else:
                 raise
-        fp = cls.open(path, mode='w')
-        fp.write(content)
-        fp.close()
-        return fp
+        with cls.open(path, mode='w') as fp:
+            fp.write(content)
+            return fp
```
stevenle: i dunno if any place in the code uses the returned `fp` but doesn't seem to make sense to return anything at all if the `fp` is closed since it would need to be re-opened anyway.

grow/rendering/rendered_document.py
```diff
@@ -23,6 +23,13 @@ def _get_tmp_filename(self):
         """Temp filename if has temporary dir."""
         return os.path.join(self.tmp_dir, self.hash)
 
+    @property
+    def filename(self):
```
stevenle: `filename` always feels like the file's basename to me, e.g. `index.html`. maybe use `local_filepath` or `local_file_path` instead?

grow/pods/storage/file_storage.py
Zoramite: I agree, I'll remove it and see if it causes any issues.

grow/rendering/rendered_document.py
Zoramite: Good point, I've switched it to `file_path` instead.

PR #624, opened 2017-11-01
Check for and use .nvmrc node version

angusm:
During 'grow install' the pod is checked for the presence of a .nvmrc file.
If one is present a subprocess sources nvm.sh and uses nvm to install the appropriate version.
During pre-processing the same approach is used to prefix the gulp command with a call to use the appropriate node version.


angusm:
Sorry, hold on.
Got messy version control on this. Working to correct it now, haven't done a proper git flow pull request for, getting my bearings, bear with me.

PR #626, opened 2017-11-01
Check for and use nvmrc file

angusm:
If an .nvmrc file is present during 'grow install' nvm is used to install the appropriate version of node. During 'grow run' if gulp is being used, 'nvm use' is run to pick up the appropriate version of node from the .nvmrc file.

### DIFF
--- a/grow/common/sdk_utils.py
+++ b/grow/common/sdk_utils.py
@@ -142,6 +142,11 @@ def get_popen_args(pod):
 
 
 def has_nvmrc(pod):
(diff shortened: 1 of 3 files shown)


angusm:
Unit tests are missing and need to be added. I wanted to check I was on the right track first before testing bad code.

Zoramite:
This looks good for now. I had an idea for improving the entire install process, there is a lot of repetition in there. So I'll merge this in, then rewrite how the install works and then I'll ping you to write some unit tests for it :)

PR #633, opened 2017-11-04
RC Based Default Flag Values

### LINKED ISSUE #600: Support default flag options in growrc
Just noticed the PR that adds support for `.growrc.yaml`. Would be cool to support default values for flags, e.g.:

```
run:
  no-ui: true
  no-update-check: true
  port: 8088
```

would be equivalent to `grow run --no-ui --no-update-check --port=8088`

Zoramite:
Adding support for rc file support to set the default flag values for CLI.

Configuration is in this format:

```yaml
grow:
  run:
    port: 8822
```

Which then changes the default for `grow run`:

```sh
               Pod: /Users/someone/somewhere
           Address: http://localhost:8822/
      Server ready. Press ctrl-c to quit.
```

Will look for rc configuration files in two places: `~/.growrc.yaml` and `{cwd}/.growrc.yaml`.

The defaults in the working directory will be used over the home directory based configuration. This allows for per project defaults.

Fixes #600


stevenle:
I did a quick pass over the code and it looks good to me as far as I can tell. Thank you, great change!

PR #640, opened 2017-11-09
Add --no-include-obsolete option to download_translations command

jeremydw:
Add --include-obsolete flag to download_translations command. The idea is we should be able to download translations from Sheets without including translations that might be obsolete/no longer used in the build. The translations can continue to live in Google Sheets if they're needed in the future (and can be removed with the --prune flag), but downloading with --no-include-obsolete should be the best way to download only what's needed.

@Zoramite do you think this should default to `False` when downloading from Google Sheets? That way, the regular usage is to only download what's needed.

### DIFF
--- a/grow/commands/subcommands/download_translations.py
+++ b/grow/commands/subcommands/download_translations.py
@@ -16,11 +16,12 @@ CFG = rc_config.RC_CONFIG.prefixed('grow.download_translations')
 @shared.locale_option(help_text='Which locale(s) to download. If unspecified,'
                                 ' translations for all locales will be downloaded.')
 @shared.service_option
-def download_translations(pod_path, locale, service):
+@shared.include_obsolete_option(CFG, default_value=True)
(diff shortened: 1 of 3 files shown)


Zoramite:
I like the idea of only download the strings that are in use in the current branch.

grow/commands/subcommands/download_translations.py
```diff
@@ -16,7 +16,7 @@
 @shared.locale_option(help_text='Which locale(s) to download. If unspecified,'
                                 ' translations for all locales will be downloaded.')
 @shared.service_option
-@shared.include_obsolete_option(CFG, default_value=True)
+@shared.include_obsolete_option(CFG, default_value=False)
```
Zoramite: Also, the default `default_value` is False :)

jeremydw: Thanks, removed the unnecessary code. :)

PR #642, opened 2017-11-10
Adding a secondary sort order by pod_path.

### LINKED ISSUE #612: Default ordering of g.docs()
I noticed that the returned docs from `g.docs(<collection>)` returns a list of documents in reverse alphabetical order by filename. Doesn't it make more sense to use alphabetical ordering?

Zoramite:
Fixes #612

### DIFF
--- a/grow/pods/collection.py
+++ b/grow/pods/collection.py
@@ -263,8 +263,11 @@ class Collection(object):
     def list_docs(self, order_by=None, locale=utils.SENTINEL, reverse=None,
                   include_hidden=False, recursive=True, inject=False):
         reverse = False if reverse is None else reverse
-        order_by = 'order' if order_by is None else order_by
-        key = operator.attrgetter(order_by, 'pod_path')
+        if order_by is None:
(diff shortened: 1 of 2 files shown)


grow/pods/collection.py
```diff
@@ -263,8 +263,11 @@ def list_categories(self):
     def list_docs(self, order_by=None, locale=utils.SENTINEL, reverse=None,
                   include_hidden=False, recursive=True, inject=False):
         reverse = False if reverse is None else reverse
-        order_by = 'order' if order_by is None else order_by
-        key = operator.attrgetter(order_by)
+        if order_by is None:
+            order_by = ('order', 'pod_path')
```
stevenle: assuming that `pod_path` is the full path relative to the pod root, the collection path prefix will always be the same, so might be better to order by the `doc.base`, which would have less chars for the str compare.

Zoramite: It does, but I believe that collections can have sub directories that are part of the collection as well? If so then the basename would mess up the sorting between the sub directories if we just use the base.

PR #648, opened 2017-11-10
Ignoring format errors when a list based yaml file is changed.

Zoramite:


jeremydw:
This will silence the error from the log, but I wonder if it's worth exploring a little more. For example, it would still be useful to see an error message in the log if indeed the document was formatted incorrectly. Maybe a larger fix for this would be to exclude list-based documents from the route tree / servable documents?

Zoramite:
Good point I'll go deeper in and ignore it when making the servable list.

jeremydw:
Nice. :) That will actually help a lot since it should be completely OK to use list-based content for partials (e.g. for lists of locales).

Separate note: it might also be good to support extracting tagged strings from list-based content too.

PR #701, opened 2017-12-11
Adding custom gettext to handle more formatting options.

### LINKED ISSUE #698: Support new Python string formatters in gettext tag
Adding this as an enhancement request to start some discussion on this. 

The old-style Python string formatters is very error prone for translators since the final `s` in `%(var)s` always looks like a mistake. The new-style Python formatters are much more clear IMO, e.g. `Hello, {name}!`.

It'd be cool if the gettext tag, `{{_(text, foo='a', bar='b')}}` could detect whether the string is using the new-style placeholders or the old-style one whenever there are placeholder variables used. 

It'd also be nice if it could do this on a string-by-string basis instead of being a setting on the whole project so that it'd be easier for old projects to transition to the new format.

Zoramite:
Supports using `{name}` in strings to make it easier for translations.

Also supports the normal python new style formatting:

ex: `{name:^10}` pads and centers the value for a 10 width.
ex: `{pi:06.2f}` formats the number -> `003.14`

Keeps the unknown keywords to prevent issues with existing behavior.

ex: `_('{name}: {value}', name='foo')` -> `foo: {value}`

See [pyformat.info](https://pyformat.info/) for more formatting.

Fixes #698

### DIFF
--- a/grow/common/structures.py
+++ b/grow/common/structures.py
@@ -10,6 +10,13 @@ class AttributeDict(dict):
     __setattr__ = dict.__setitem__
 
 
+class SafeDict(dict):
+    """Keeps the unmatched format params in place."""
+
+    def __missing__(self, key):
+        return '{' + key + '}'
+
+
 class SortedCollection(object):
     """Sequence sorted by a key function.
 
--- a/grow/common/utils.py
+++ b/grow/common/utils.py
@@ -9,6 +9,7 @@ import json
 import logging
 import os
 import re
+import string
 import sys
 import threading
 import time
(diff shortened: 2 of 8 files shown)


stevenle:
I might not be the best to review the actual code since I'm not familiar with how the original code works, but could you verify that the old format continues to work? And could you add unit tests that has more than one placeholder (e.g. `hello {name1} and {name2}`) and an example of a translation that mixes them up (e.g. `hola {name2} y {name1}`)?

Zoramite:
@stevenle I've added a test that will check the rendered output with multiple placeholders and used the tagged syntax to simulate different locales.

PR #741, opened 2018-02-27
Updating dependencies.

Zoramite:


Zoramite:
I hope so too, I've tried to keep track of the dependencies that mess up things when they get updated.

PR #771, opened 2018-05-19
Allow for custom HTTP headers in S3 deployments

willthomson:
Functionality mimics what is already present for Google Cloud Storage
deployments. Headers can be specified in the podspec.yaml in the format
shown in grow.io’s podspec.yaml file:

https://github.com/grow/grow.io/blob/master/podspec.yaml


Zoramite:
Looks good, thanks for the PR!

PR #791, opened 2018-07-10
Fix issue where None items inside dependency graph caused log spam.

jeremydw:
Not sure if this is the right fix; I think we might want to avoid adding `None` to the dependency graph in the first place.


Zoramite:
Yeah, `None` should not be in the dep graph. I'll convert this PR to fix the issue when adding to the dep graph.

PR #808, opened 2018-08-08
Adding check to ignore columns without a grid key.

Zoramite:


grow/preprocessors/google_drive.py
```diff
@@ -257,6 +257,8 @@ def download(cls, spreadsheet_id, gids=None, format_as='list', logger=None,
                             row = row[1:]
                             row_len = len(row)
                             for col, grid_key in enumerate(headers):
+                                if not grid_key:
```
jeremydw: Wonder if we should also exclude keys that start with `_`? This is opinionated but it might be helpful? Not sure.

Zoramite: Good idea. Updated.

PR #823, opened 2018-08-24
Install yaml-dev in alpine docker image

stevenle:


stevenle:
I tested this by running:

```
docker run -it grow/base:alpine-latest
```

Then ran the following:

```
grow --version  # got the "missing libyaml" message
apk add --update yaml-dev
pip install --force-reinstall --update pyyaml
grow --version  # no more "missing libyaml" message
```

PR #830, opened 2018-09-04
Ignoring the node modules from the server filewatcher.

Zoramite:
Reduces the amount of watched files by a  *lot*.

### DIFF
--- a/grow/preprocessors/file_watchers.py
+++ b/grow/preprocessors/file_watchers.py
@@ -48,6 +48,8 @@ class PreprocessorEventHandler(events.PatternMatchingEventHandler):
 
     def __init__(self, preprocessor, *args, **kwargs):
         self.preprocessor = preprocessor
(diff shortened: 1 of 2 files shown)


stevenle:
what files are tracked by the file watcher? wondering if we should add more items to that list.

Zoramite:
Currently *all* files are watched by the watcher... are there others that we should add?

stevenle:
i dunno how it's used in general, but maybe the modules that are creating the file watcher objects should send over a list of includes to watch. watching an entire project dir seems like a waste of resources if the only real directory you need to watch is `/content/`, for example.

stevenle:
anyway, i think this change is probably good to go and sometime in the future you can maybe consider refactoring this module a bit so that there's more structure/transparency about which modules are watching which files. 

only other file that might be worth adding to the ignore right now is `/build/`.

PR #833, opened 2018-09-06
Subprocess Killing improvements

Zoramite:
Improve how the subprocesses are killed after the main process has exited. This should prevent the preprocessor from persisting as a separate process.

### DIFF
--- a/grow/preprocessors/gulp.py
+++ b/grow/preprocessors/gulp.py
@@ -64,12 +64,15 @@ def _kill_child_process():
             pass
 
     # Give the process some time to finish by itself.
-    time.sleep(3)
 
-    for process in _child_processes:
-        try:
+    while True:
+        time.sleep(1)
+
+        is_pending = False
+        for process in _child_processes:
             if process.poll() is None:
+                is_pending = True
                 process.kill()
-        except OSError:
-            # Ignore errors.
-            pass
+
+        if not is_pending:
+            return
--- a/grow/preprocessors/gulp.py
+++ b/grow/preprocessors/gulp.py
@@ -3,6 +3,7 @@
 import os
 import atexit
 import subprocess
+import time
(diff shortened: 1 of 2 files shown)


grow/server/manager.py
```diff
@@ -83,6 +83,7 @@ def start(pod, host=None, port=None, open_browser=False, debug=False,
             serving.run_simple(host, port, app, request_handler=handler, threaded=True)
             done = True
         except socket.error as e:
+            # if any(x in str(e) for x in ('Errno 48', 'Errno 98')):
```
stevenle: you should maybe do a regular for loop here since this'll evaluate each of the expressions even if the first one returns `true`

Zoramite: I agree that it makes it evaluate each string check, but looking at how the other logic would need to change, it makes it more complicated to split it into a normal loop.

grow/preprocessors/gulp.py
```diff
@@ -58,7 +59,17 @@ def _kill_child_process():
     for process in _child_processes:
         try:
             process.terminate()
-            process.wait()
         except OSError:
             # Ignore the error.  The OSError doesn't seem to be documented(?)
             pass
+
+    # Give the process some time to finish by itself.
+    time.sleep(3)
```
stevenle: it looks like you're always enforcing a 3-second lag on every shutdown, even if the child process takes less time than that to shut down. i'd suggest sleeping on some smaller interval and checking if the process is still running, and force killing it after the max of 3 seconds.

grow/server/manager.py
stevenle: alright, since N is always going to be small here, i don't mind this pattern here, but in general we should prefer writing optimal code where possible, otherwise small memory or cpu issues can seem trivial at first but snowball into a mess in the future.

PR #846, opened 2018-10-02
Turning off the old school formatting based on the pod config.

### LINKED ISSUE #822: Feature request: podspec setting to only support one style of string formatting
The current gettext tag supports both styles of formatting:

https://github.com/grow/grow/blob/master/grow/templates/tags.py
```
        value = value % variables
        value = utils.safe_format(value, **variables)
```

It'd be cool if we could enforce one style or another using a podspec setting.

At the moment, for projects that use only the new-style Python formatting with `{placeholders}`, any percent symbol will error because of the `value = value % variables` line.

Zoramite:
Fixes #822

### DIFF
--- a/grow/templates/tags.py
+++ b/grow/templates/tags.py
@@ -171,13 +171,17 @@ def make_doc_gettext(doc):
     return gettext
 
 
-def make_gettext(func):
+def make_gettext(pod, func):
     """Create a gettext function that tracks translation stats."""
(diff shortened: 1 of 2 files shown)


stevenle:
i think this looks fine to me, but i wonder if we should do a major version increase (`0.6.0`?) since it's a breaking change.

Zoramite:
Yeah, that was my thought. I guess I haven't updated the milestones yet for that... I'll get them updated.

PR #857, opened 2018-10-18
Removing old style routing from Grow

Zoramite:


jeremydw:
I'm comfortable removing the old routing once #870 is fixed – since there's no workaround for #870 once the old routing is removed. :)

PR #909, opened 2019-04-01
Adding a filter config for the routing to be able to filter routes.

### LINKED ISSUE #902: Ability to ignore paths to build as part of deploy
When building larger projects it can be useful to have builds ignore specific collections or paths while building.

This can be partially done a bit with using flags during the build to only build certain paths or with tagged paths that don't build in specific environments, but that is an incomplete solution. If you tag a document to have a blank path in specific environments then it cannot be linked to (ex `doc.path`) since the doc would have a blank path.

One use case for this feature is for projects that have a lot of automated pages (such as a blog or preprocessor that syncs many pages). When the number of pages being imported increases the amount of time to build the site increases. By using a deploy specific ignore functionality a project can build only the relevant pages but not mess up any links (like in headers/footers). The subset of files could then be deployed separately from the normal site files (ex: separate service in GAE). This speeds up the normal builds since they can ignore the bulk pages and only build the parts they are responsible for without messing up the links and other features while links to the ignored docs still work correctly.

Zoramite:
Fixes #902


matthiasrohmer:
Isn't that fixing #902? 🤔

Zoramite:
@matthiasrohmer Thanks for the catch. Updated it to the correct issue.

PR #911, opened 2019-04-01
Updating to write pod cache files even if there are rendering errors.

### LINKED ISSUE #904: Build error handling
When running `grow build`, if any template error occurs, grow continues to try to build the entire site before failing. This can take a lot of time for large sites. Can we stop the builder at the first known build error?

Also, doesn't seem like objectcache is being saved to file when builds err. For sites that use google_cloud_image, that means we need to re-ping the service again for all images.

Zoramite:
Fixes #904


stevenle:
👍

PR #927, opened 2019-05-08
Adding a filter for string formatting using the newer string format

Zoramite:
Normal Jinja `format` filter: `{{ "%s - %s"|format("Hello?", "Foo!") }}`
New `format_str` filter: `{{ "{} - {}"|format_str("Hello?", "Foo!") }}`


Zoramite:
This was not needed, can do `'something {}'.format()` and it works like normal.

PR #933, opened 2019-05-24
Fixing issue with dev routing that causes paths with no base file to have conflicts.

Zoramite:
Adding a clean collection path for the document that strips out the locale.

### DIFF
--- a/grow/documents/document.py
+++ b/grow/documents/document.py
@@ -169,15 +169,15 @@ class Document(object):
     @property
     def collection_base_path(self):
         """The base directory inside the collection."""
-        return self.collection_path[:-len(self.basename)]
+        return self.collection_sub_path[:-len(self.basename)]
 
     @property
-    def collection_path(self):
+    def collection_sub_path(self):
         """The pod path relative to the collection path."""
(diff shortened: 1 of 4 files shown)


grow/documents/document.py
```diff
@@ -176,6 +176,12 @@ def collection_path(self):
         """The pod path relative to the collection path."""
         return self.pod_path[len(self.collection.pod_path):]
 
+    @property
+    def collection_path_clean(self):
```
stevenle: Is "collection_path" the best name? It feels like "collection_path" would refer to the collection's path, i.e. `/content/pages/` in the tests below.

Zoramite: I'm open to other suggestions. `collection_sub_path`? I'm not sure what else to call it. Basically it is the pod path minus the collection path.

so in the `/content/pages/` collection a `pod_path` of `/content/pages/sub/index.yaml` would be `/sub/index.yaml`. It is the path inside of the collection... but I wasn't sure what to call it.

Zoramite: Renamed it to `collection_sub_path` I think that matches it better.

PR #940, opened 2019-06-06
Switch out the slugify with a library

### LINKED ISSUE #939: Slugify should escape apostrophes
Slugify is used in some extensions (like the xml_feed preprocessor) that generate filenames. We should exclude characters like quotes/apostrophes from being included in slugified names.

Zoramite:
Changing how slugify works to use the slugify library instead of the built-in method.

Can use the legacy slugify for now by adding this to your `podspec.yaml`:

```
features:
  legacy_slugify: true
```

Fixes #939

### DIFF
--- a/grow/documents/document.py
+++ b/grow/documents/document.py
@@ -1,11 +1,11 @@
 """Grow documents."""
 
-import datetime
 import json
 import logging
 import os
 import re
 import yaml
+from slugify import slugify
 from grow.common import structures
 from grow.common import untag
 from grow.common import urls
@@ -398,7 +398,11 @@ class Document(object):
         value = self.fields.get('$slug')
         if value:
             return value
-        return utils.slugify(self.title) if self.title is not None else None
+        if not self.title:
+            return None
+        if self.pod.features(self.pod.FEATURE_OLD_SLUGIFY):
+            return utils.slugify(self.title)
+        return slugify(self.title)
 
     @property
     def sitemap(self):
--- a/grow/documents/document.py
+++ b/grow/documents/document.py
@@ -4,8 +4,8 @@ import json
 import logging
 import os
 import re
+import slugify
 import yaml
-from slugify import slugify
 from grow.common import structures
 from grow.common import untag
 from grow.common import urls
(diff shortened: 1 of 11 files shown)


grow/documents/document.py
```diff
@@ -1,11 +1,11 @@
 """Grow documents."""
 
-import datetime
 import json
 import logging
 import os
 import re
 import yaml
+from slugify import slugify
```
stevenle: If we're following the google style guide, we should import the module here (`import slugify`) instead of the function within the module (`from slugify import slugify`)

grow/pods/pods.py
```diff
@@ -158,6 +163,18 @@ def _load_extensions(self, load_local_extensions=True):
             self._extensions_controller.register_extensions(
                 self.yaml.get('ext', []))
 
+    def _load_features(self):
+        config = self.yaml.get('features', {})
+        for key, value in config.iteritems():
+            # Features can be turned on with a True value.
+            # But the true does does not act as a configuration.
```
stevenle: Not sure I understand what this comment is saying, could you re-phrase?

stevenle: Also, can we make the legacy string placeholders use the same enable/disable as what you're creating here?

Zoramite: Yeah, we can make the legacy placeholder use it too, I'll do that in a separate pull request though.

grow/documents/document.py
Zoramite: Updated to use just `import slugify`

grow/pods/pods.py
Zoramite: I've also updated the phrasing here os it is hopefully easier to understand.

grow/pods/pods.py
```diff
@@ -144,8 +149,8 @@ def _get_bytecode_cache(self):
     def _load_experiments(self):
         config = self.yaml.get('experiments', {})
         for key, value in config.iteritems():
-            # Experiments can be turned on with a True value.
-            # But the true does does not act as a configuration.
+            # Features can be turned on with a True value
```
Zoramite: I'm going to a separate pull request to clean up this process of reading from a config into the features object so it doesn't need to be part of the pod.

PR #957, opened 2019-06-20
Fix an issue with caching and localized yaml.

jeremydw:


Zoramite:
I need to look into this one a bit more. It is following the same logic that we use in the pod to read yamls: https://github.com/grow/grow/blob/master/grow/pods/pods.py#L846-L858

The idea is that it reads the raw, tagged version of the file, caches that, then does the untagging and caches the untagged version with the locale.

I'll do some deeper looking to see where this goes wrong and make sure that this is the correct fix for the issue.

PR #973, opened 2019-06-27
Updating install script to install to a binary directory.

### LINKED ISSUE #962: Installer should suggest adding $HOME/bin to $PATH instead of using alias
CLIs in general should avoid installation by adding `alias` lines to bash configuration files. I think it's better that grow installs to `$HOME/bin` and suggest adding `export PATH=$PATH:$PATH/bin` to their bashrc (or equivalent).

Zoramite:
Also checks the path to make sure the directory is on the PATH.

Fixes #962

### DIFF
--- a/install.py
+++ b/install.py
@@ -17,15 +17,16 @@ import zipfile
 
 DOWNLOAD_URL_FORMAT = 'https://github.com/grow/grow/releases/download/{version}/{name}'
 RELEASES_API = 'https://api.github.com/repos/grow/grow/releases'
-ALIAS_FILES_LINUX = ['.bash_aliases', '.bash_profile', '.profile', '.bashrc']
-ALIAS_FILES_MAC = ['.bash_profile']
-ALIAS_FILES_DEFAULT = '.bashrc'
-ALIAS_RE = re.compile(r'^alias grow\=([\'"])(.*)\1$', re.MULTILINE)
+RC_FILES_LINUX = ['.bash_profile', '.profile', '.bashrc']
+RC_FILES_MAC = ['.bash_profile']
+RC_FILE_DEFAULT = '.bashrc'
 
 if 'Linux' in platform.system():
     PLATFORM = 'linux'
+    BIN_PATH = '/usr/local/bin/grow'
 elif 'Darwin' in platform.system():
     PLATFORM = 'mac'
+    BIN_PATH = '/usr/local/bin/grow'
 else:
     print('{} is not a supported platform. Please file an issue at '
           'https://github.com/grow/grow/issues'.format(sys.platform))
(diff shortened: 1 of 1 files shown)


stevenle:
So IMO we should try to preserve backwards compatibility for those who currently use the previous method for updating their `grow` binary, since their `alias` setting would override the new `PATH`. I'm thinking we detect whether an `alias` for grow is set and try to see if they're using the old download location or not.

install.py
```diff
@@ -17,15 +17,16 @@
...
-ALIAS_FILES_MAC = ['.bash_profile']
-ALIAS_FILES_DEFAULT = '.bashrc'
-ALIAS_RE = re.compile(r'^alias grow\=([\'"])(.*)\1$', re.MULTILINE)
+RC_FILES_LINUX = ['.bash_profile', '.profile', '.bashrc']
+RC_FILES_MAC = ['.bash_profile']
+RC_FILE_DEFAULT = '.bashrc'
 
 if 'Linux' in platform.system():
     PLATFORM = 'linux'
+    BIN_PATH = '/usr/local/bin/grow'
```
stevenle: The issue with using /usr/local/bin/ is that it requires `sudo` to write to that directory. I'd recommend keeping the old `$HOME/bin` location and having the user include that in their $PATH.

install.py
```diff
@@ -17,15 +17,16 @@
 
 DOWNLOAD_URL_FORMAT = 'https://github.com/grow/grow/releases/download/{version}/{name}'
 RELEASES_API = 'https://api.github.com/repos/grow/grow/releases'
-ALIAS_FILES_LINUX = ['.bash_aliases', '.bash_profile', '.profile', '.bashrc']
-ALIAS_FILES_MAC = ['.bash_profile']
-ALIAS_FILES_DEFAULT = '.bashrc'
-ALIAS_RE = re.compile(r'^alias grow\=([\'"])(.*)\1$', re.MULTILINE)
+RC_FILES_LINUX = ['.bash_profile', '.profile', '.bashrc']
```
stevenle: I think the fallback order should be `['.bashrc', '.bash_profile', '.profile']`, and should be the same for both Linux and Mac. We might also want to look into adding `zsh` config locations here as well since the next major version of Mac will use `zsh` as default.

install.py
```diff
@@ -104,7 +75,13 @@ def get_release_for_platform(releases, platform):
     return None
 
 
+def has_bin_in_path(bin_path):
+    """Determine if the binary path is part of the system paths."""
+    return '{}:'.format(bin_path) in os.environ['PATH']
```
stevenle: Hmm, this wouldn't resolve to true if they set their path like:

```
export PATH=$PATH:$HOME/bin
```

stevenle: Maybe something like:
```
return bin_path in os.environ['PATH'].split(':')
```

Zoramite: Thanks, updated.

install.py
stevenle: I think having a single known location would be easier to maintain, and I think I generally like `$HOME/bin` for installers that are automatically installed via `curl ... | bash`, since it's less scary.

Zoramite: I've updated it to only install to the `~/bin` by default on all OSes and does the path append for those that don't have that directory in the `PATH` already.

install.py
Zoramite: Updated the ordering and added the zsh config file as one of the places to search.

PR #993, opened 2019-08-17
Add deployment option to inspect routes command

jeremydw:


grow/commands/subcommands/convert.py
```diff
@@ -19,11 +19,15 @@
 
 @click.command()
 @shared.pod_path_argument
+@shared.deployment_option(CFG)
 @click.option('--type', 'convert_type', type=click.Choice(CONVERT_CHOICES))
-def convert(pod_path, convert_type):
+def convert(pod_path, convert_type, deployment):
```
Zoramite: Why does the convert command need the deployment?

jeremydw: I think it's because the generated `routes.json` also depend on environment settings, which  aren't pulled over from the YAML tagging. For example,

```
$path: /foo/
$path@env.prod: /bar/
```

Will result in different `routes.json` depending on the deployment

Zoramite: Yeah, but the `convert` command is for updating files between versions of grow, not for parsing environment specific settings... so I was just confused.

jeremydw: Oh, I use it with the `collection_routing` conversion type. :)

Zoramite: Ok, sounds good.

PR #994, opened 2019-08-19
Remove PyCrypto Dep and update other dependencies.

### LINKED ISSUE #991: pycrypto?
Hi,
pycrypto has latest release in 2014 and I can't even install it via pip on Void Linux as it uses LibreSSL. pycrypto does not exist in OpenBSD as well.

PyCryptodome should be modern alternative - https://www.pycryptodome.org/en/latest/src/introduction.html

Zoramite:
Fixing the dependencies to no longer require pycrypto. Also updating dependencies that can be updated.

Fixes #991


jeremydw:
I'd love to remove the PyCrypto dep in Grow; it's a big bloater of our image. When I tried this last I ran into some issue; I can't remember what that was but if it's safe to remove now that would be great.

Zoramite:
I'm going to try it out. I think we had to keep with one of the other things that we removed, but I'm not sure at this point. I don't see any github issues that relate to anything with pycrypto missing.

PR #1010, opened 2019-09-27
Doesn't error out on empty page but already fetched posts

### LINKED ISSUE #1009: Blogger pagination breaks at multiples of 10
Problem: Blogger pagination breaks when page is full, but the next page has no items.

Fix: incoming - once I can make a build and run tests.

JaredSartin:
Fixes #1009 

EDIT: The problem is that the blogger API sends a next key when the page is full (10 posts), regardless of the total amount of items. The next page came back with no results, which invoked the error/


Zoramite:
Can you rebase this on the latest grow? it looks like it is using an older branch?

Zoramite:
There are some extra things in this PR that I don't think are supposed to be there. I cleaned up what I think is the key part in #1045

PR #1056, opened 2020-02-18
Add Java to Docker image for use with Soy/Closure Templates extension

jeremydw:
Need to test this prior to merging to ensure there are no install prompts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,9 @@ RUN apt-get update \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+RUN add-apt-repository ppa:linuxuprising/java
+RUN apt-get install oracle-java10-installer oracle-java10-set-default
+
 # Node, Yarn, GCloud sources.
 RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - \
   && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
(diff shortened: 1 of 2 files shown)


stevenle:
I don't think this is a good idea. You can always create another docker image that inherits from the base grow image and install all of the extra dependencies that way. I don't think base grow necessarily needs to support Java since that's a lot of extra overhead.

jeremydw:
Fair point – let's create a new image so projects that require Java can use it. What's the right convention here – `Dockerfile.java`? And, do we need to set up something else in Docker Hub to make this work?

stevenle:
Do you need it as part of the core grow project? I feel like this is easy enough to do directly within the project folder that requires it instead of publishing to a public image.

But if you need a convention, you can maybe use the one that google cloud sdk uses:
https://github.com/GoogleCloudPlatform/cloud-sdk-docker

From their docs:

```
google/cloud-sdk:latest, google/cloud-sdk:VERSION: (large image with additional components pre-installed, Debian-based)

google/cloud-sdk:slim, google/cloud-sdk:VERSION-slim: (smaller image with no components pre-installed, Debian-based)

google/cloud-sdk:alpine, google/cloud-sdk:VERSION-alpine: (smallest image with no additional components installed, Alpine-based)
```

jeremydw:
Abandoning this in favor of a project-local `Dockerfile` built with Google Cloud Build. If we have >1 project or a Grow extension requiring Java we can provide the `Dockerfile` and build instructions there.

PR #1070, opened 2020-04-13
Removing the Translation Injection

jeremydw:
NOTE: Not ready to merge.

This is just a checkpoint for a refactor to move the translations downloader and uploader to a Google API batch request. (This is how the strings preprocessor works today.)


jeremydw:
@Zoramite - Sorry to ask, can you pick this up from me? I'd like to solve three things:

1. Remove the `inject` feature and related code paths for translations. This was never used in practice, and with the Live Editor its need will be entirely nullified.
2. Change the `grow translations download` command to use `batchGet` (similar to how the `strings` preprocessor works.
3. Change the `grow translations upload` command to do the same batches.

This should:

- Eliminate the need for the threading code related to the `translations download|upload` commands
- Make the commands work much more quickly

Zoramite:
Converting this PR to just remove the injection. Will follow up with the PR for adding batch downloads.

PR #1103, opened 2020-06-16
Removal of pyinstaller dependency and packaged apps.

Zoramite:


jeremydw:
There are other dependencies that we've added I believe specifically for PyInstaller compatibility:

- certifi
- pyasn1
- six
- pytz

I don't see any imports in our codebase directly related to those packages.

Note that these may still be included transitively but I think we can remove them from our Pipfile. Not sure if you want to do those in a separate PR or combine them here as part of this PR.

PR #1121, opened 2020-10-01
Improved in docs 

eklare19:


jeremydw:
Hi! Thanks for sending a PR. Unfortunately, these grammar and wording nits don't make much sense. I'd prefer to leave the docs as-is. Short, terse, and sweet is preferred.

PR #1150, opened 2021-04-22
Use temp file when writing objectcache. Fixes #1147

### LINKED ISSUE #1147: Objectcache reading/writing with multiple shards
For large sites that build with sharding enabled, there's currently a race condition where one shard may be writing to a large objectcache file while another shard is attempting to read it, causing a JSON parse error.

For writing the objectcache, can we do one of two things?

1) Use a multiprocesslock
2) Write to a tmp file (e.g. `objectcache.json.tmp`) and then rename it

jeremydw:


Zoramite (approved):

stevenle:
this doesn't quite fix it since two processes can still try to write to that file at the same time. can you try something like `tempfile.mkstemp()` ?

jeremydw:
> this doesn't quite fix it since two processes can still try to write to that file at the same time. can you try something like `tempfile.mkstemp()` ?

Yes. 🤦

PR #1152, opened 2021-04-22
Use temp file when writing objectcache. Fixes #1147

### LINKED ISSUE #1147: Objectcache reading/writing with multiple shards

jeremydw:


grow/cache/podcache.py
```diff
@@ -105,7 +106,7 @@ def _write_json(self, path, obj):
         output = json.dumps(
             obj, cls=json_encoder.GrowJSONEncoder,
             sort_keys=True, indent=2, separators=(',', ': '))
-        temp_path = '{}.tmp'.format(path)
+        _, temp_path = tempfile.mkstemp()
```
stevenle, line 109: the first return is an open file descriptor, can you use that to write the file directly instead of using the pod's write_file wrapper?

something like:

```
fp, temp_path = tempfile.mkstemp()
fp.write(output)
fp.close()
# then move the file
```

grow/cache/podcache.py
```diff
@@ -105,9 +107,10 @@ def _write_json(self, path, obj):
         output = json.dumps(
             obj, cls=json_encoder.GrowJSONEncoder,
             sort_keys=True, indent=2, separators=(',', ': '))
-        temp_path = '{}.tmp'.format(path)
-        self._pod.write_file(temp_path, output)
-        self._pod.move_file_to(temp_path, path)
+        fp, temp_path = tempfile.mkstemp()
+        fp.write(output)
+        fp.close()
+        os.rename(temp_path, path)
```
stevenle, line 113: can you verify that `path` here is the full filesystem path and not the pod path?

jeremydw: Updated to use `pod.abs_path`.

jeremydw:
OK, this should be working now with the standard methods. Agree that it's probably best to not modify code paths as much as possible, in order to reduce coupling across the codebase.

grow/cache/podcache.py
```diff
@@ -105,9 +108,14 @@ def _write_json(self, path, obj):
         output = json.dumps(
             obj, cls=json_encoder.GrowJSONEncoder,
             sort_keys=True, indent=2, separators=(',', ': '))
-        temp_path = '{}.tmp'.format(path)
-        self._pod.write_file(temp_path, output)
-        self._pod.move_file_to(temp_path, path)
+        _, temp_path = tempfile.mkstemp()
+        with open(temp_path, 'w') as fp:
```
stevenle, line 112: Somewhat of a micro-optimization but it's better to open the file descriptor since it's already available to the process.

```
fd, path = tempfile.mkstemp()
with os.fdopen(fd, 'w') as fp:
  fp.write(output)
```

(you also won't need to make sure the dir exists with this since it's guaranteed to be writeable)

jeremydw: Updated; but flagging the "makedirs" is for the "real path", not the "temp path". In testing it looks like the real path can be a folder that doesn't exist when this is invoked (since it might go into the `.grow` folder).

grow/cache/podcache.py
```diff
@@ -105,9 +108,14 @@ def _write_json(self, path, obj):
...
-        self._pod.write_file(temp_path, output)
-        self._pod.move_file_to(temp_path, path)
+        fd, temp_path = tempfile.mkstemp()
+        with os.fdopen(fd, 'w') as fp:
+            fp.write(output)
+            real_path = self._pod.abs_path(path)
+            dir_name = os.path.dirname(real_path)
+            if not os.path.exists(dir_name):
+                os.makedirs(dir_name)
+            shutil.copyfile(temp_path, real_path)
```
stevenle, line 118: I actually think what you had before was more correct (`os.rename`?). `copyfile` will transfer bytes from one location to another, so afaik you'll still have the issue with one process reading in the middle of another process writing. I think `os.rename` will just update a pointer location which doesn't have the bytes being copied and will make it less likely for there to be conflicts.

stevenle: Also you should move the stuff after the `fp.write(output)` line outside of the `with` statement.

jeremydw: OK - I think this is in its final form now. I was using shutil before because I thought mkstemp deleted the file outside the with statement. That was wrong (the file sticks around), so I can use os.rename.

stevenle (approved):

PR #1153, opened 2021-04-22
Ability for pod to write file using temp files.

Zoramite:
Adding an argument to the storage class for using a temp file instead of a direct write to a file.


stevenle:
i think given that our plan is to eventually switch projects to amagaki, i'd think i'd favor https://github.com/grow/grow/pull/1152 more for its simplicity. afaik there aren't a lot of places where there would be race conditions in file writes, just the objectcache file.

PR #1203, opened 2022-04-20
fix: add instructions for using standard virtualenvs

lgiordani:
Some developers might prefer to manage virtual environments the traditional way, or to use other virtual environment managers like pyenv (which I use). I think giving instructions on how to set up the development environment in those cases is cheap and doesn't damage the standard setup of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,38 @@ pipenv run grow install
 pipenv run grow run
 ```
 
+## Development without pipenv
+
+If you are using standard virtual environments or a different virtual environment manager (like pyenv) you can follwo this instructions.
+
+### Install libyaml
+
+```bash
+# On Mac with Homebrew (https://brew.sh/).
+brew install pipenv libyaml
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -43,6 +43,38 @@ pipenv run grow install
 pipenv run grow run
 ```
 
+## Development without pipenv
+
+If you are using standard virtual environments or a different virtual environment manager (like pyenv) you can follwo this instructions.
```
jeremydw, line 48: Fix typos; also, could you move this to a CONTRIBUTING.md if this is meant for Grow developers (not users)?

lgiordani:
I moved it to `CONTRIBUTING.md` but I kept part of it in `README.md` as it applies also to users (it's the way I installed it). Can you check again? Thanks!

lgiordani:
@jeremydw Can you have a look at this and at the other outstanding PRs? Thanks!

PR #1204, opened 2022-04-20
fix: add instructions to test using pytest

lgiordani:
This adds instructions on how to test Grow using pytest. As the framework is compatible with `Unittest` the whole suite works out of the box. For the reasons behind this addition see https://github.com/grow/grow/discussions/1199

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,20 @@ pipenv run grow install
 pipenv run grow run
 ```
 
+## Testing with pytest
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -43,6 +43,20 @@ pipenv run grow install
 pipenv run grow run
 ```
 
+## Testing with pytest
```
jeremydw, line 46: As per the other issue, move this to `CONTRIBUTING.md`

